PR #7, opened 2020-03-15
wasp: pinetime: Release SPI NOR from deep power-down

### LINKED ISSUE #6: SPI NOR not responding to RDID, RDSR or RDCR
PineTime does not reset to the watch face after DFU micropython.zip via NRFConnect Android app. Used GNU-RM as recommended in README.md.

After DFU, I can use wasptool --console to enter REPL and enter contents of main.py manually - which displays the watch face. I can then `wasp.run()` and Ctrl-X to keep it running - but If I long-press the button to return to boot loader, it will no longer respond to a 2 second press to get back into micropython. This means I must DFU to reset the watch.

wasptool --exec does not work. When using --exec, the tool completes without errors, but no code runs on the watch and a subsequent --console enters the REPL where it presents a prompt (nothing executed).

I tried a known good micropython.zip someone else built and it too exhibits the same behaviors as my builds.

salcedo:
Fixes #6

### DIFF
--- a/wasp/boards/pinetime/watch.py
+++ b/wasp/boards/pinetime/watch.py
@@ -65,6 +65,12 @@ i2c = I2C(1, scl='I2C_SCL', sda='I2C_SDA')
 touch = CST816S(i2c)
 vibrator = Vibrator(Pin('MOTOR', Pin.OUT, value=0), active_low=True)
 
+# Release flash from deep power-down
+nor_cs = Pin('NOR_CS', Pin.OUT, value=1)
+nor_cs(0)
+spi.write('\xAB')
+nor_cs(1)
+
 # Mount the filesystem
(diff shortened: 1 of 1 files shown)


daniel-thompson:
Well spotted. First let me discuss with @peterhinch since I think it would probably make sense to integrate wakeup code (and some PM support) into the FLASH class itself.

daniel-thompson:
Sorry, this has been a while. Whilst I'd like to eventually move this to the flash driver itself, it is better to have this fix than not.

Also, I finally sorted out proper license information for wasp-os and I think this PR pre-dates that. Are you OK to contribute this under GNU LGPLv3+?

salcedo:
Good to merge. Thanks for asking.

PR #10, opened 2020-03-19
Fixed the documentation for setting the RTC time

jmwright:
As far as I can tell, `set_time` hasn't existed since this commit: https://github.com/daniel-thompson/wasp-os/commit/127df66335657cbd77539a509d507a4fcfb403b2

I haven't been able to track down when it disappeared, but it doesn't exist in the current RTC driver. The original implementation didn't allow for setting the date anyway it seems. This PR updates the readme to reference `set_localtime()` instead.

I couldn't get `set_time()` to work when working through the readme, but I'm new to wasp-os within the last couple of days. Feel free to just close this if I've understood something fundamental.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ To set the time and restart the main application:
 
 ~~~
 ^C
-watch.rtc.set_time((hh, mm, ss))
+watch.rtc.set_localtime((yyyy, mm, dd, HH, MM, SS))
 wasp.run()
 ~~~
 
--- a/README.md
+++ b/README.md
@@ -88,6 +88,12 @@ watch.rtc.set_localtime((yyyy, mm, dd, HH, MM, SS))
 wasp.run()
 ~~~
 
+Or just use:
+~~~
+./tools/wasptool --rtc
+~~~
+which can run these commands automatically.
+
 As mentioned above there are many drivers and features still to be
 developed, see the [TODO list](TODO.md) for current status.
 


daniel-thompson:
This change is entirely correct: set_time() was removed at the point proper calendar code was enabled.

Having said that it would also be good to add: " (or just use `./tools/wasptool --rtc` which can run these commands automatically):" to the line above.

jmwright:
@daniel-thompson I've added that. Let me know what you think.

daniel-thompson:
Looks good.

Incidentally, I realized that I haven't yet pushed a license to cover the wasp-os specific code (micropython and the bootloader already have clear licenses). I'm planning to adopt GNU LGPLv3 or later. This is OK for your contributions?

jmwright:
> I'm planning to adopt GNU LGPLv3 or later. This is OK for your contributions?

That's fine with me.

daniel-thompson:
Merged. Thanks!

jmwright:
Thanks for wasp-os. I'm wearing my PineTime right now, and will probably seal the back within a couple of days and start wearing it full time. Having access to the REPL and being able to push updates to it via BLE are huge advantages.

daniel-thompson:
No worries. It's great to see the interest in wasp-os. Note that there are some big incompatible changes comes for the apps fairly soon but that will hopefully be worth it because it will give the apps a nice way to interact with the manager. That doesn't mean don't write apps... just expect to have to make a few changes as we evolve the interfaces!

It exciting to see people gluing up their watches based on this work.

I haven't been able to do that yet. I wear it daily but I'm using strong double sided tape (which isn't water resistant). Basically I want to have a tested way to to update the bootloader (and softdevice) before I seal mine (and once I have it tested I will start sharing known working bootloader binaries to reduce the risk of bricking).

jmwright:
> Note that there are some big incompatible changes comes for the apps fairly soon but that will hopefully be worth it because it will give the apps a nice way to interact with the manager.

That sounds good, and I had kind of gotten that sense from the roadmap. If I create any apps before then, I'll be prepared to port them over to the new way of doing things. Just for reference, he's what I (tentatively) have in my mind for the first changes/features/apps I'm interested in doing. We'll see how much time I have to devote to them.

- Simple change to the way the date is displayed on the clock (baby step into wasp development).
- Alarm feature with weekly repeat (may not have all the UI parts to set it, may just be set through the REPL)
- Accelerometer based wake - Some sort of gesture based functionality to turn the screen on without having to hit the side button. The pre-installed PineTime firmware has this, but it didn't work all that well for me.
- Filtered notifications - This is just in the early stages of thought right now, but I really just want the notifications (text, email, calendar) that matter to me without all the distracting ones that can wait. This will probably have some form of filters (sender, keywords, etc), and may be an app on a Raspberry Pi that hits multiple APIs to aggregate notifications, then sends those to the watch somehow.

PR #34, opened 2020-06-11
Update preprocess to use $(BOARD)

SteveAmor:
The preprocess to create watch.py was hard coded for PineTime so changed to $(BOARD)
This will be required for other watches like the P8

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,7 @@ reloader: bootloader
 softdevice:
 	micropython/ports/nrf/drivers/bluetooth/download_ble_stack.sh
 
-micropython: wasp/boards/pinetime/watch.py
+micropython: wasp/boards/$(BOARD)/watch.py
 	$(MAKE) -C micropython/mpy-cross
 	$(RM) micropython/ports/nrf/build-$(BOARD)-s132/frozen_content.c
 	$(MAKE) -C micropython/ports/nrf \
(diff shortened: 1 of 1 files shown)


SteveAmor:
Is it ok to use my @users.noreply.github.com email address in the Signed-off-by: tag?

SteveAmor:
I'm now not sure if this is the best way to address this issue for the P8 watch as the files in ```wasp/boards/pinetime/``` are identical to the files that you would put in ```wasp/boards/p8/``` which adds duplication to development.  Also, if further variants of watches are added, that adds to the duplication.

daniel-thompson:
The current makefile is clear wrong therefore I am happy to take the change.

Eventually however most of pinetime/watch.py will need to be extracted into library code so it can be used by multiple boards and that bit of the Makefile will change again.

SteveAmor:
Did I update the PR correctly?

SteveAmor:
Signed-off-by: tag added.

daniel-thompson:
Sorry to keep you waiting... the commit hasn't got square brackets round it. I was planning to fix that up myself but it means I can't merge via the web interface so I haven't got round to it yet (mostly because I've been focussing on getting data off the hrs3300 instead).

I'll get to it this week.

daniel-thompson:
Applied. Thanks.

PR #36, opened 2020-06-13
Adds P8 'button on right' png for bootloader

SteveAmor:


daniel-thompson:
I tweaked the commit to align it with the bootloader updates and after that... applied. Thanks!

PR #38, opened 2020-06-20
Update Readme to run the simulator

SteveAmor:
Signed-of-by: Steve Amor <steveamor@users.noreply.github.com>


daniel-thompson:
I decided I'd rather fix the bug in the Makefile ;-). See 64afca12 ("Makefile: Fix make sim (allow BOARD to be unset)").

PR #45, opened 2020-07-25
Fix typo

mscherer:


daniel-thompson:
Thanks for the fix. Could you change the title to include the area being modified ("README: fix typo") and add a Signed-of-by: block in the commit.

The Signed-of-by: is an explicit statement that your contribution comes under a, b, c, or d from the following list: https://developercertificate.org/

PR #48, opened 2020-07-31
Fix background color of strings.

Siroj42:
Before, the background color was only applied to the direct background of the characters, as seen in this picture. 
Bildschirmfoto von 2020-07-31 13-51-20

The padding and the spacing between the characters was not affected. After these changes, it now looks as expected when using a custom background color:
Bildschirmfoto von 2020-07-31 14-34-11

### DIFF
--- a/wasp/draw565.py
+++ b/wasp/draw565.py
@@ -291,21 +291,23 @@ class Draw565(object):
         display = self._display
         bgfg = self._bgfg
         font = self._font
+        bg = self._bgfg >> 16
 
         if width:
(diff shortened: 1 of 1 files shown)


daniel-thompson:
Thanks for the fix. Could you change the first line of the description of each patch to include the area being modified ("draw565: Fix...") and add a Signed-off-by: block within the description.

The Signed-off-by: (formatted as "Signed-off-by: Your Name <your@email>") is an explicit statement that your contribution comes under a, b, c, or d from the following list: https://developercertificate.org/

PR #52, opened 2020-08-01
Add GitHub Actions workflow for building wasp-os artifacts

MirkoCovizzi:
This workflow is manually triggerable. An example of the artifacts generated can be found [here](https://github.com/MirkoCovizzi/wasp-os/actions/runs/191287244). The whole workflow takes about 4 minutes to complete. These artifacts can be automatically downloaded for future publishing/releasing workflows ([by using the GitHub Action `actions/download-artifact@v2`](https://docs.github.com/en/actions/configuring-and-managing-workflows/persisting-workflow-data-using-artifacts)). 

They can also be manually downloaded and, after extraction, they can be flashed on the respective smartwatches.

In the future it is possible to add testing workflows and trigger them on PRs and other actions.

### DIFF
--- a/.github/workflows/wasp-os.yml
+++ b/.github/workflows/wasp-os.yml
@@ -1,0 +1,41 @@
+name: wasp-os
+
+on:
+  workflow_dispatch:
+    inputs:
+      logLevel:
+        description: 'Log level'
+        required: true
+        default: 'warning'
+      tags:
+        description: 'Build firmware'
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        board: [pinetime, p8]
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Install dependencies
+        run: sudo apt-get update -y && sudo apt-get install git build-essential
+          libsdl2-2.0.0 python3-click python3-numpy python3-pexpect python3-pil python3-pip python3-serial
+      - name: Install Python dependencies
+        run: pip3 install --user pysdl2
+      - name: Setup toolchain
+        run: sudo apt-get update -y && sudo apt-get install gcc-arm-none-eabi
+      - name: Checkout wasp-os
+        uses: actions/checkout@v2
+      - name: Setup submodules
+        run: make submodules
+      - name: Setup softdevice
+        run: make softdevice
+      - name: Build firmware
+        run: make -j `nproc` BOARD=${{ matrix.board }} all
+      - name: Persist artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ matrix.board }}-artifact
+          path: |
+            bootloader-daflasher.zip
+            micropython.zip


AlwaysLivid:
Hey, I also just opened a pull request (just 3 minutes after you!) for building wasp-os artifacts everytime a commit is made.

https://github.com/daniel-thompson/wasp-os/pull/53

However, my workflow cannot be manually triggered and is specific to the PineTime. I was thinking whether you'd want me to invite you to my own branch of `wasp-os`, so that we can figure out a way to integrate both of our (similar, but still different!) changes simultaneously.

MirkoCovizzi:
@AlwaysLivid I think that at the moment building artifacts for every commit or pull request is a bit too much. We have first to figure out together with @daniel-thompson if he wants to have GitHub Actions workflows and for doing what. At the moment I feel the priority should be easing the releases, so, if you want, we can work together and make a workflow for that, but first we need to hear what @daniel-thompson has to say about it.

AlwaysLivid:
> @AlwaysLivid I think that at the moment building artifacts for every commit or pull request is a bit too much/frequent (it would be better to have that on specific tags). We have first to figure out together with @daniel-thompson if he wants to have GitHub Actions workflows and for doing what. I feel the priority right now should be easing the releases, so, if you want, we can work together and make a workflow for that, but first we need to hear what @daniel-thompson has to say about it.

The fact is that it's commonplace for [other firmwares that are available on the PineTime](https://twitter.com/MisterTechBlog/status/1289577296072830981). Bleeding-edge builds don't take away the importance of stable releases that could be later distributed via official channels, but they're still useful for people. I personally went through all of this trouble to learn how to use GitHub Actions because I've been struggling with the `arm-none-eabi` toolchain for over 3 weeks straight now.

As far as our implementations are concerned, there are two important differences;
- Your implementation can be manually triggered, which would be a function that would be particularly useful, considering that my own implementation caches both the toolchain and the Python modules. Caching saves a lot of resources and time-- at the expense of theoretical, yet probable, compilation errors due to outdated or otherwise problematic Python dependencies.
- My implementation only tests against the PineTime and **not** for the P8.

daniel-thompson:
> Either way, I'd also be interested to know what @daniel-thompson thinks about this.

This post is a little declarative but hopefully that makes it clear.

My thoughts are that CI actions are useful and should run on every commit. I believe that should also have the side effect of automatically compile testing PRs for me.

If we use CI to automatically test PRs then it would also be useful for CI actions to fail if the top commit does not have a Signed-off-by: tag to show that the contributor has signed of a, b, c, or d from the Developer Certificate of Origin (https://developercertificate.org/ ).

CI should build for all supported devices (although maybe this would have to change if we end up supporting a very large number of devices).

PS Note that if you are both cooperating on the final patch then whilst the patch can only have one Author in the git metadata it can have multiple Signed-off-by: to indicate authorship and you could also use the "Co-developed-by: Their Name <their@email>" tag that is getting more popular with Linux kernel developers to share credit.

Ideally there should only be one artefact and it should be structured similar to a real binary release containing binaries for all supported devices. I've started to move in that direction by moving micropython.zip and friends into build--$(BOARD)/ directories. However at present I haven't added a bindist target to construct a full release using the Makefile.

AlwaysLivid:
> My thoughts are that CI actions are useful and should run on every commit. I believe that should also have the side effect of automatically compile testing PRs for me.

Cool, that's in progress.

> If we use CI to automatically test PRs then it would also be useful for CI actions to fail if the top commit does not have a Signed-off-by: tag to show that the contributor has signed of a, b, c, or d from the Developer Certificate of Origin (https://developercertificate.org/ ).

There's a GitHub Action that allows that. I will promptly add that to my own pull request.

```
     - name: Verify that the commit has been signed off
       uses: vchain-us/verify-action@master
```

> CI should build for all supported devices (although maybe this would have to change if we end up supporting a very large number of devices).

> PS Note that if you are both cooperating on the final patch then whilst the patch can only have one Author in the git metadata it can have multiple Signed-off-by: to indicate authorship and you could also use the "Co-developed-by: Their Name [their@email](mailto:their@email)" tag that is getting more popular with Linux kernel developers to share credit.

Isn't it `Co-authored-by`?

> Ideally there should only be one artifact and it should be structured similar to a real binary release containing binaries for all supported devices. I've started to move in that direction by moving micropython.zip and friends into build--$(BOARD)/ directories. However at present I haven't added a bindist target to construct a full release using the Makefile.

So, does it need to follow the following structure?

```
Artifact (.zip archive) -> P8/PineTime folders -> micropython.zip, bootloader.hex, bootloader-daflasher.zip, reloader.zip
```

daniel-thompson:
> There's a GitHub Action that allows that. I will promptly add that to my own pull request.

:+1: 

> > PS Note that if you are both cooperating on the final patch then whilst the patch can only have one Author in the git metadata it can have multiple Signed-off-by: to indicate authorship and you could also use the "Co-developed-by: Their Name [their@email](mailto:their@email)" tag that is getting more popular with Linux kernel developers to share credit.
> 
> Isn't it `Co-authored-by`?

The kernel uses co-developed-by (https://www.kernel.org/doc/html/latest/process/submitting-patches.html#when-to-use-acked-by-cc-and-co-developed-by ) however the text unconstrained so it doesn't matter to much. Using Co-developed-by might occasionally allow automated tools to see further though.

> > Ideally there should only be one artifact and it should be structured similar to a real binary release containing binaries for all supported devices. I've started to move in that direction by moving micropython.zip and friends into build--$(BOARD)/ directories. However at present I haven't added a bindist target to construct a full release using the Makefile.
> 
> So, the structure needs to look like this, right?
> 
> ```
> Artifact (.zip archive) -> P8/PineTime folders -> micropython.zip, bootloader.hex, bootloader-daflasher.zip, reloader.zip
> ```

More or less... it also needs some of the tools included (wasptool and friends for example).

Essentially it should look like wasp-os-0.2(-remastered).tar.gz: 
https://github.com/daniel-thompson/wasp-os/releases/tag/v0.2

Ignore the remastered BTW... there was a mistake in the original packaging so I hope not to have to remaster the 0.3 release.

> I have prepared some sort of a draft that should work, but I'd also like to talk with @MirkoCovizzi about this.

Yes, I'm interested to see how this turns out. Note that I will be away from keyboard for most of this week but I'll take a look at whatever emerges when I get back.

daniel-thompson:
> I have prepared some sort of a draft that should work, but I'd also like to talk with @MirkoCovizzi about this.

Was consensus reached about which approach is the best way forward here (I will make a call if needed but I'd prefer consensus ;-) ).

AlwaysLivid:
I haven't been able to reach out to Mirko. @daniel-thompson

MirkoCovizzi:
@AlwaysLivid Hi, sorry, I was on holiday today. Reach me over to Discord, we can talk there if you want.

daniel-thompson:
This topic went quiet so I've combined the code from #53 with a few ideas from this (and other places) into #97. Thanks to all who have discussed this.

PR #53, opened 2020-08-01
Added GitHub Actions support

panos:
- This workflow caches both the Python modules, as well as the arm-none-eabi toolchain itself, in order to save resources and time.
- This workflow is meant to imitate the building instructions as closely as possible for every board.
- After the workflow compiles wasp-os for a specific board, it will package the reloader, the bootloader (both the standalone and the DaFlasher variations), a copy of MicroPython, the documentation and the included tools in individual .tar.gz archives right afterwards.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,86 @@
+name: CI for wasp-os
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        board: [pinetime, p8]
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout files
+      id:   checkout-files
+      uses: actions/checkout@v2
+
+    - name: Check cached Python modules
+      id:   cache-pip
+      uses: actions/cache@v2
+      with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
+          restore-keys: |
+              ${{ runner.os }}-pip-
+
+    - name: Check the cached arm-none-eabi-gcc compiler
+      id:   cache-toolchain
+      uses: actions/cache@v2
+      env:
+        cache-name: cache-toolchain
+      with:
+          path: ${{ runner.temp }}/arm-none-eabi
+          key:  ${{ runner.os }}-build-${{ env.cache-name }}
+          restore-keys: ${{ runner.os }}-build-${{ env.cache-name }}
+
+    - name: Install arm-none-eabi-gcc
+      id:   install-toolchain
+      # installs arm-none-eabi if the CI environment can't find it in the cache
+      if:   steps.cache-toolchain.outputs.cache-hit != 'true'
+      uses: fiam/arm-none-eabi-gcc@v1.0.2
+      with:
+          release: 9-2019-q4
+          directory: ${{ runner.temp }}/arm-none-eabi
+
+    - name: Make submodules
+      id:   make-submodules
+      run:  |
+          export PATH=$PATH:${{ runner.temp }}/arm-none-eabi/bin
+          make -j BOARD=${{ matrix.board }} submodules
+    
+    - name: Make softdevice
+      id:   make-softdevice
+      run:  |  
+          export PATH=$PATH:${{ runner.temp }}/arm-none-eabi/bin
+          make -j BOARD=${{ matrix.board }} softdevice
+
+    - name: Build firmware
+      id:   make-firmware
+      run:  |
+          export PATH=$PATH:${{ runner.temp }}/arm-none-eabi/bin
+          make -j BOARD=${{ matrix.board }} all
+
+    - name: Create archive
+      id:   create-archive
+      run:  |
+          tar -cJf wasp-os-${{ github.sha }}-${{ matrix.board }}.tar.gz \
+          build-${{ matrix.board }} \
+          COPYING \
+          COPYING.LGPL \
+          README.rst \
+          TODO.rst \
+          tools \
+          docs
+
+    - name: Upload build
+      id:   upload-firmware
+      uses: actions/upload-artifact@v2
+      with:
+          name: wasp-os-${{ github.sha }}-${{ matrix.board }}.tar.gz
+          path: wasp-os-${{ github.sha }}-${{ matrix.board }}.tar.gz
+


AlwaysLivid:
This pull request and #52 show similarities-- with a few key differences that could work well together. I've turned this pull request into a draft until it's clear what needs to happen from this point on. (There's an ongoing discussion in the aforementioned pull request)

AlwaysLivid:
So, I ended up modifying my GitHub Actions workflow in a way that mostly aligns with what you requested. I have yet to squash my commits right now, but before merging, I'd just like to know whether the current file structure is alright, or if I'm also supposed to bundle everything in a single `.tar.xz` file. I'm not sure how I can bundle all of these files inside of a folder called `wasp-os-{{ commit revision }}`, but I can try! :D

See the latest build log [here](https://github.com/AlwaysLivid/wasp-os/actions/runs/192986633).

AlwaysLivid:
@daniel-thompson As I'm going to be busy over the next few days, I'd like to ask if the Makefile now output all files to `build-${{ matrix.board }}` by itself? If that's the case, then I'd need to remove the following lines from `make-firmware`:

```
          mkdir build-${{ matrix.board }}
          mv bootloader.hex build-${{ matrix.board }}
          mv bootloader-daflasher.zip build-${{ matrix.board }}
          mv micropython.zip build-${{ matrix.board }}
          mv reloader.zip build-${{ matrix.board }}
```

AlwaysLivid:
This took me quite a bit, but I modified the workflow so that it puts everything in `.tar.gz` archives, nearly precisely as requested, alongside with the README files, the copyright notices, the documentation and the tools. I also had to add the `BOARD` variable to the `make submodules` and `make softdevice`-- I believe that this was intended. The only difference (I hope) in contrary to what I was requested to do was separating the archives per board, but I figured that this would keep things simple for everyone-- and adding new boards only takes a single line change!

I followed a naming format that's similar to the one used in the first release (`wasp-os-[full SHA1 hash]-[board].tar.gz`).

I did originally add a sign-off checker, but the platform that was being used was proprietary and had an uneasy amount of quirks (e.g. it breaks on force-pushes). I hope all of that is fine-- if there are any other changes that need to be made, do let me know.

AlwaysLivid:
> Looks good overall. Unfortunately I can't figure out how the pip cache gets populated in the first place. Does it rely on an additional file somehwere?

Not sure what you mean exactly. After downloading all the packages using your `Makefile` configuration, the workflow "stashes" the installed packages and retains them for a single week, if I recall correctly.

daniel-thompson:
> > Looks good overall. Unfortunately I can't figure out how the pip cache gets populated in the first place. Does it rely on an additional file somehwere?
> 
> Not sure what you mean exactly. After downloading all the packages using your `Makefile` configuration, the workflow "stashes" the installed packages and retains them for a single week, if I recall correctly.

I saw the cache... but I can't see where the patch the cache gets populated (there is no build step the checks for a hit on cache-pip .

> I presume that I need to sign it off?

Yes please. Line breaks in the bulletted lists in the description would be good too.

AlwaysLivid:
> I saw the cache... but I can't see where the patch the cache gets populated (there is no build step the checks for a hit on cache-pip .

I'm seriously unsure as to what exactly you mean. If there's a cached version of the packages, that means that the packages are already there, so pip won't have to redownload everything from scratch. Nothing further needs to be done, other than caching that specific folder (for the next time, when the build runs again).

AlwaysLivid:
I hope that the commit message is fine. I can still change it if necessary.

daniel-thompson:
> > I saw the cache... but I can't see where the patch the cache gets populated (there is no build step the checks for a hit on cache-pip .
> 
> I'm seriously unsure as to what exactly you mean. If there's a cached version of the packages, that means that the packages are already there, so pip won't have to redownload everything from scratch. Nothing further needs to be done, other than caching that specific folder (for the next time, when the build runs again).

I'm not worried about the case when the cache exists. What I didn't understand is, if the cache is cold, how does the build rule know what packages to install?

Having looked a bit closer I'm beginning to realized the wildcard might be able to pick up requirements.txt file from within some of the submodules. To be honest I'm a little worried about that since it will be installing modules that aren't mentioned in the README, especially so if the wildcard ends up picking up reloader/lib/CMSIS/CMSIS/NN/Tests/UnitTest/requirements.txt .

AlwaysLivid:
> I'm not worried about the case when the cache exists. What I didn't understand is, if the cache is cold, how does the build rule know what packages to install?
> 
> Having looked a bit closer I'm beginning to realized the wildcard might be able to pick up requirements.txt file from within some of the submodules. To be honest I'm a little worried about that since it will be installing modules that aren't mentioned in the README, especially so if the wildcard ends up picking up reloader/lib/CMSIS/CMSIS/NN/Tests/UnitTest/requirements.txt.

The workflow is meant to replicate whatever you would do on your personal machine. I haven't used any `requirements.txt` file, I just consciously followed what the documentation said line by line-- the mentioned make commands **install Python packages**. You can review one of the build logs in my own repositories as well to confirm that. The workflow saves the cache folder. That means that nothing will probably happen the next time the `Makefile` is going to attempt to install the packages, unless if there's an update. It just keeps them in there, so that there won't be any resources wasted on downloading the same Python packages over and over again.

daniel-thompson:
> I haven't used any `requirements.txt` file, I just consciously followed what the documentation said line by line-- the mentioned make commands **install Python packages**.

They do? Are you quite sure? I certainly didn't intend them to do anything like that!

> You can review one of the build logs in my own repositories as well to confirm that. The workflow saves the cache folder. That means that nothing will probably happen the next time the `Makefile` is going to attempt to install the packages, unless if there's an update. It just keeps them in there, so that there won't be any resources wasted on downloading the same Python packages over and over again.

Can you share a direct link? All the jobs I can find either have no cache or restore a 22 byte (e.g. empty) cache.

AlwaysLivid:
Nevermind, I am not quite sure and I'm seriously starting to doubt what I said earlier. Sorry about that. There seem to be a few other issues with my implementations, but I'm being helped out by Mirko. I will get back later.

AlwaysLivid:
@daniel-thompson Me and @MirkoCovizzi were sharing our concerns as to whether `make submodules` and `make softdevice` should take a `BOARD=` variable, with the fact that we're talking about fetching singular git submodules in every possible case here considered. I perceived that as some method of futureproofing, but I'm not sure that the bigger picture is astoundingly clear apart from that.

daniel-thompson:
@AlwaysLivid, @MirkoCovizzi: It's a bug. I found in independantly early this afternoon (`make sim` is showing the same problem). I hope to fix it shortly... for now maybe use `make BOARD=FIXME softdevice` in the actions files).

daniel-thompson:
> @AlwaysLivid, @MirkoCovizzi: It's a bug. I found in independantly early this afternoon (`make sim` is showing the same problem). I hope to fix it shortly... for now maybe use `make BOARD=FIXME softdevice` in the actions files).

This should be fixed in 28961f1 .

daniel-thompson:
Applied (via #97).

PR #57, opened 2020-08-05
apps: gameoflife: Added to P8/PineTime builds

panos:
I believe that shipping Daniel Thompson's implementation of Conway's Game of Life by default for the time being, instead of providing the user with a simplistic version of the firmware, will help with illustrating the capabilities of wasp-os in a way that's clear and tangible. There's already a video that does illustrate how the Game of Life looks on the PineTime watch (see: https://www.youtube.com/watch?v=YktiGUSRJB4), but I am confident that a filmed demonstration is inferior compared to a hands-on example. I strongly believe that the inclusion of this small game could very well promote adoption, as well as, to an extent, further contributions, without the end user having to go through the alienating hassle of understanding how the codebase works in order to try it out by themself.


AlwaysLivid:
> Enabling the game by default was in the plan... but only as part of fixing the TODO on line 116.

I'd like to look into it-- I cannot seem to spot the TO-DO line that was brought up, however.

daniel-thompson:
> I'd like to look into it-- I cannot seem to spot the TO-DO line that was brought up, however.

Sorry I did rather neglect the file name... the TODO is in wasp.py (just before we register the apps) to say that we are registering the apps in totally the wrong place and need to move it into user configurable code.

However, how we handle user configureable code is still a bit funky. The concept is easy... the user can modify main.py however they like and this is where the apps ought to be registered. It would allow Game of Life to be easily disabled to save memory. The problem comes when we have to make an incompatible API change (wasp-os is still young... it will happen) and need to update the default main.py.

It is an area that causes support problems because users are not yet trained to handle main.py updates. This especially affects users who *haven't* actually changed main.py. One option I've been toying with is to simply have a list of checksums of unmodified main.py files... that way if we recognise the checksum we know the user never changed their main.py is ok to update it (I'm less worried about users who do tweak main.py because they probably already know how to update it).

daniel-thompson:
The RAM use used by Game of Life to record the board state when the app is not running is too high to enable this by default. It doesn't cause end-user problems by it does make it more difficult for users to write their own applications. Since making writing applications easy is the "one pervasive goal [of wasp-os] that influences almost everything about it, from its name to its development roadmap" then I'm still unwill to enable this by default.

However... there have been a few changes.

1. The Game of Life *is* enabled automatically in the simulator (where the the RAM limit is measured in gigabytes rather than kilobytes)
2. The Game of Life is [documented in the application library](https://wasp-os.readthedocs.io/en/latest/apps.html#conway-s-game-of-life) with a hint about how to enable it.

There is still scope for more documentation about how to really make use of `main.py` but for now this what we are doing to raise awarenesss of the game with users.

PR #59, opened 2020-08-06
fixing a minor typo on readme.rst

karl01101101:


daniel-thompson:
Thanks for the patch. This looks like a duplicate of #45... although it you're happy to act on https://github.com/daniel-thompson/wasp-os/pull/45#issuecomment-667660999 then I guess it doesn't matter which one is pulled.

PR #66, opened 2020-08-20
Remove attempt to wake flash at this level

wbjeffrey:
Wake from deep power down is now handled in the driver. Remove attempt to wake the chip from this file.

Signed-off-by: Jeffrey Bailey <wb.jeffrey@gmail.com>

### DIFF
--- a/wasp/boards/pinetime/watch.py.in
+++ b/wasp/boards/pinetime/watch.py.in
@@ -96,15 +96,9 @@ try:
                     _callback)
     vibrator = Vibrator(Pin('MOTOR', Pin.OUT, value=0), active_low=True)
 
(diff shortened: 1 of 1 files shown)


daniel-thompson:
Sorry for dragging my feet with this (has it really been three weeks... seems like no time at all).

AFAICT its a great contribution and I'm really pleased to see it (so thanks very much... and please do not be discouraged). However since it moves everything into common code I need to test it on three differentt devices and I haven't yet found the time. I'll be on it soon!

wbjeffrey:
No problem, I understand.

   --Jeff

daniel-thompson:
Merged. Thanks.

PR #72, opened 2020-09-17
Update install.rst

Jac-eko:
Add apt command to install GNU-RM toolchain
I'm not sure if there is too much packages, but it works

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -44,6 +44,13 @@ tested using the `GNU-RM toolchain
 <https://developer.arm.com/tools-and-software/open-source-software/developer-tools/gnu-toolchain/gnu-rm>`_
 (9-2019-q4) from Arm.
 
+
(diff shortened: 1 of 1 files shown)


panos:
> Add apt command to install GNU-RM toolchain
> I'm not sure if there is too much packages, but it works

I think that could confuse people using other operating systems, unless if custom commands were to be written for as many case scenarios as possible.

daniel-thompson:
Thanks for the PR although I confess I am a little reluctant to accept it.

The cross-compiler toolsets included with the distros are not the same as the Arm GNU-RM toolchain. The reason I explicitly indicate that all build and test is done using the Arm GNU-RM toolchain is that in the past there have been some significant problems with arm-none-eabi toolchains on some distros, espeically when building floating point code.

In other words if the Debian Buster toolchain works then great but it's not what we normally build and test with and, for that reason, is not the recommended toolchain.

PR #74, opened 2020-10-01
Adding a new clock application

tht-jxny:
This pull request is to introduce a new type of clock to the quick ring of wasp-os. The clock is called a Fibonacci Clock. I know, somewhat nerdy B)
### How to tell the time:
The Fibonacci sequence is a sequence of numbers created by the Italian mathematician Fibonacci in the 13th century. This is a sequence starting with 1 and 1, where each subsequent number is the sum of the previous two. For the clock I used the first 5 terms: 1, 1, 2, 3 and 5.

The screen of the clock is made up of five squares whose side lengths match the first five Fibonacci numbers: 1, 1, 2, 3 and 5. The hours are displayed using red and the minutes using green. When a square is used to display both the hours and minutes it turns blue. White squares are ignored.

To tell time on the Fibonacci clock you need to do some math. To read the hour, simply add up the corresponding values of the red and blue squares. To read the minutes, do the same with the green and blue squares. The minutes are displayed in 5 minute increments (0 to 12) so you have to multiply your result by 5 to get the actual number.

If wanted, I could make some modifications so that each and every minute can be displayed.

And here is how it looks - at least on the simulator:
fibo_clock

I also added a new vertical Battery Meter so my new watchface won't iterfere with the battery.
Feedback is appreciated!


daniel-thompson:
Thanks for the PR.

I shall do a line-by-line review in a moment but two things to say upfront.

1. Really cool idea. I have a spare PineTime to gift to a family member shortly... they will really like this watch face so some version of this will definitely land in the repo so they can get access to it.

2. I don't like addinng an alternative battery meter, especially so when there is lots of empty space below the date. In particular there are other icons that should appear in the top row of most clocks (currently the BT and notification icons) and these as currnetly designed to balance against the "weight" of the vertical meter.

daniel-thompson (changes requested):

tht-jxny:
I hope that the new commit is in concert with your expectations. I don't know but if one really wants to save RAM, one has to go over the code again and refactor several elements. I just don't know if this would matter that much.

daniel-thompson:
This is now merged: https://github.com/daniel-thompson/wasp-os/commit/4942e3e93533821dcb4cd96cebac0ea9d5ffe8ce

I did a few small cleanups to get a clean history (the merge soup in the PR requiring rather too many fixups for my taste).

daniel-thompson:
Ah... forgot to explain the "dropped manifest changes". I didn't like the way adding an app requires changes to three files so I unified them in a later patch: https://github.com/daniel-thompson/wasp-os/commit/c1c66fb487113acbd4b2b7a722c3782d5a962e08

With the latest mainline your clock will be in the flash by default but, as discussed, not active unless the user adds it to their `main.py`.

PR #76, opened 2020-10-04
Added borders to apps in the launcher

Kozova1:
I added borders between the apps using 2 lines, one horizontal and one vertical.  
I also added a settings page to configure them, similar to the color test.

### DIFF
--- a/wasp/apps/launcher.py
+++ b/wasp/apps/launcher.py
@@ -74,6 +74,8 @@ class LauncherApp():
 
     def _draw(self):
         """Redraw the display from scratch."""
+        border_color = wasp.system.launcher_border_color
+
         def draw_app(app, x, y):
             if not app:
                 return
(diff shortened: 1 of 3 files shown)


Kozova1:
I also added a color block similar to the one that's seen in the color test app, to visualize the color.

daniel-thompson:
Thanks for the PR.

However I'm afraid to say I'm not really very keen on adding borders to the launcher. There isn't really much prior art (e.g. UI examples from other products) for adding lines to icon grids and I don't think it boosts useability. That doesn't mean you can't have it but for something so much tailored to personal taste should not, I think, be adding to the master repo.

On the other hand... I think this **is** a great thing to add to the manual as an example of customizations that can be installed into the spinor filesystem using wasptool. Essentially your draw code can be added by sub-classing the launcher and overriding the draw method. After that it is a simple change to main.py to swap out the stock launcher with the customized one.

wasp-os' filessytem (and main.py) is designed to allow customization without having to recompile the whole firmware but this is not currently very well explained in the manual so this could be a great example of the sophisticated tricks that can be deployed.

PR #81, opened 2020-10-15
Set RTC with nRF Connect

Carglglz:
This PR adds a `Current Time Service` and `Date Time Characteristic` to the BLE profile, so connecting to the watch through nRF Connect allows to set the `watch.rtc`  local time.

This is done by writing in `Date Time Characteristic` the local time in hex format.

I included a script `tools/hexlocaltime.py` to get this easily. 
Connect to the device (will appear as `wasp-ble`) then do:
```
wasp-os$ python tools/hexlocaltime.py
Set localtime: 2020-10-15 16:46:40 --> Hex: E4070A0F102E28
```
 Write ` E4070A0F102E28`  in Date Time Characteristic and send it.
Now the watch should display the updated local time.

PD: I also included a vibration pulse when the watch receives a BLE connection/disconnection event.

PD2: I tried to include a Battery Service/Battery Level too but It seems it allows only one Service/Characteristic (maybe not enough memory...)

[EDIT]

I'm working on alternative tools to `pynus.py` for `tools/wasptool --console`  and `./tools/wasptool --rtc` based on [bleak](https://github.com/hbldh/bleak) which is cross-platform so it will allow to use Windows/MacOS for development too.

So far I've got this working on a Raspberry Pi Zero W (Buster), and MacOS too*.

* I think that there is a bug in the nrf bluetooth stack (I don't now if it is `ble_drv` or `bleak_uart`) but it suddenly disconnects after a few seconds. In devzone.nordicsemi forum there are similar threads about this [issue](https://devzone.nordicsemi.com/f/nordic-q-a/36688/nrf52840-android-disconnect-after-28-seconds-of-connection
) 
I will raise a ticket about this later 👍 

Signed-off-by: Carlos Gil <carlosgilglez@gmail.com>


Carglglz:
After a few days of testing, I wouldn't follow this path just yet. The bluetooth stack in the nrf port is not as stable as I though it would be. Trying to set the peripheral to add services and characteristics will override the NUS Service, so that means no more console/REPL access, and no GadgetBridge support.  

There are also some bluetooth events that are not handle in `ble_drv.c` which can cause the ble "handshake" to fail in some devices. This was fixed in Micropython upstream (https://github.com/micropython/micropython/commit/fc1f22a097ac2c6eb418605f36b53936e5138451)

On the other hand, the connection parameters (`BLE_MIN_CONN_INTERVAL `, `BLE_MAX_CONN_INTERVAL`) are set a 7.5 ms which are too low for MacOS (>= 15 ms), which causes the ble "handshake" to fail too.

There may be also some buffer overflow events or unhandled errors in NUS Service  which causes the device to reset (if the NUS Service is stressed, e.g. sending too much too often)

Based on this I will wait until the bluetooth stack gets more reliable. 👍

daniel-thompson:
Thanks for the update (and sorry I didn't reply before now).

Thanks also the the pointer to the updates on the master micropython branch. I have a couple of devices with really solid BLE connections (happily including both my normal phone and my normal laptop) but I have seen problems on others.

W.r.t. to the current time service/characteristic I guess if we implemented them in the C code that implement NUS then we'd get a new useful way to set the clock without breaking existing connectivity.

Carglglz:
> Thanks also the the pointer to the updates on the master micropython branch. 
👍 

There are also some bug fixes in `VFS` related to path handling and changing directories although they are evident only when using shell commands.
https://github.com/micropython/micropython/pull/6324
https://github.com/micropython/micropython/pull/6085
https://github.com/micropython/micropython/pull/6040
https://github.com/micropython/micropython/pull/6011
https://github.com/micropython/micropython/pull/6009
https://github.com/micropython/micropython/pull/6004
https://github.com/micropython/micropython/pull/5988
I did cherry-picked those changes and for now everything works fine in wasp-os.

Also I do have some minor fixes for the nrf port, Would you accept PR's in your Micropython fork?

daniel-thompson:
> Also I do have some minor fixes for the nrf port, Would you accept PR's in your Micropython fork?

It depends...

I'll certainly take fixes for the code that is unique to wasp-os (like the boards, config, etc) and I'll also consider backported fixes that you have landed with the upstream project (although I might start to more actively rebase on the micropython master branch to avoid carrying too many backports).

I'm reluctant to take "generic" fixes unless there is a corresponding PR to the "real" micropython project.

daniel-thompson:
Given current status I'm inclined to Close this for now.

It's not a "hard no" though. It's a really neat bit of code and you are very welcome to reopen if you think it can be made solid.

PR #87, opened 2020-10-21
Add Music Player with GadgetBridge support

Carglglz:
This adds a minimal music play app with GadgetBridge support:

-> Tap to play/pause
-> Swipe left/right to next/previous track
-> Swipe up/down to volume up/down

Displays on top: Artist
Displays at the bottom: Track

This will work as far as the play/pause is not too stressed (NUS Service will crash, and the device will reset).


Carglglz:
Closing this to make a clean PR

PR #88, opened 2020-10-22
app: add Music Player App with GadgetBridge support

Carglglz:
This adds a minimal music player app with GadgetBridge support:

-> Tap to play/pause
-> Swipe left/right to next/previous track
-> Swipe up/down to volume up/down

Displays on top: Artist
Displays at the bottom: Track

This will work as far as the play/pause is not stressed too much. (NUS Service will crash, and the device will reset).

Signed-off-by: Carlos Gil <carlosgilglez@gmail.com>


daniel-thompson:
I'm sorry it's taken me so long to get round to this.

I have had to devote **all** my wasp-os my time to getting OTA updates working from the new PineTime factory firmware. This has been getting fairly urgent since it was making if difficult for new users to install wasp-os ([video here](https://youtu.be/lPasAt1LJmo) if you are interested).

Anyhow, that's all behind me know and I'm going back over the pending PRs.

wasp/apps/musicplayer.py
```diff
@@ -0,0 +1,208 @@
...
+        self._center_at = int((self._max_display - self._icon_size)/2)
+        self._musicstate = 'pause'
+        self._artist = ''
+        self._track = ''
+        self._state_changed = True
+        self._track_changed = True
+        self._artist_changed = True
+        self._track_pos = 24 + 144
+        self._artist_pos = 12
+
```
daniel-thompson, line 54: This seems like a very large number of class variables. I'm rather worried about the RAM usage.

If it works there's no need to worry too much... but I'll be inclined to add this to the (optional_ Application Library rather than enabling it by default for the time being.

wasp/gadgetbridge.py
```diff
@@ -28,14 +28,17 @@
 true = True
 false = False
 
+
```
daniel-thompson, line 31: Why are all there lots of blank links being added in this file?

wasp/gadgetbridge.py
```diff
@@ -49,6 +52,10 @@ def GB(cmd):
             wasp.system.notify(id, cmd)
         elif task == 'notify-':
             wasp.system.unnotify(cmd['id'])
+        elif task == 'musicstate':
+            wasp.system.togglemusic(cmd)
```
daniel-thompson, line 56: method_names_should_use_snake_case

wasp/gadgetbridge.py
```diff
@@ -49,6 +52,10 @@ def GB(cmd):
...
+        elif task == 'musicinfo':
+            wasp.system.setmusicinfo(cmd)
```
daniel-thompson, line 58: Ditto...

wasp/icons.py
```diff
@@ -138,7 +138,8 @@
 up_arrow = (16, 9, b'\x07\x02\r\x04\x0b\x06\t\x08\x07\n\x05\x0c\x03\x0e\x01 ')
 
 # 1-bit RLE, generated from res/down_arrow.png, 17 bytes
-down_arrow = (16, 9, b'\x00 \x01\x0e\x03\x0c\x05\n\x07\x08\t\x06\x0b\x04\r\x02\x07')
+down_arrow = (
```
daniel-thompson, line 141: Similar to the comments about whitespace... it is better not to make non-functional changes to code that is not impacted by the music player code.

wasp/apps/musicplayer.py
```diff
@@ -0,0 +1,208 @@
...
+class MusicPlayerApp(object):
+    """ Music Player Controller application.
+
+    .. figure:: res/MusicPlayerApp.png
+        :width: 179
+
+        Screenshot of the Music Player application
+    """
+    NAME = 'MPlayer'
```
daniel-thompson, line 28: I think just "Music" here is better.

wasp/apps/musicplayer.py
```diff
@@ -0,0 +1,208 @@
...
+
+    def _draw_track(self):
+        """Redraw track info"""
+        if self._track:
+            draw = wasp.watch.drawable
+            chunks = draw.wrap(self._track, 240)
+            self._fill_space('down')
+            for i in range(len(chunks)-1):
+                sub = self._track[chunks[i]:chunks[i+1]].rstrip()
+                draw.string(sub, 0, self._track_pos + 24 * i, 240)
```
daniel-thompson, line 196: These two functions seem "wrong"

They basically have identical structure and are only parameterized by the variable they act on (self._track in this case) and something with a similar name (self._track_pos). Can't we have_draw_label(self, label, pos)?

wasp/wasp.py
```diff
@@ -116,18 +123,20 @@ def __init__(self):
...
-                # Let's not bring the whole device down just because there's
-                # an exception starting one of the apps...
-                pass
+             (StepCounterApp, True),
+			 (StopwatchApp, True),
+			 (HeartApp, True),
+			 (FlashlightApp, False),
+			 (SettingsApp, False),
+             (TestApp, False),
+             (MusicPlayerApp, False)):
```
daniel-thompson, line 132: As mentioned, I'd rather not enable the MusicPlayer until I've had a change to examine RAM usage... but I'm happy to merge it before I do that.

Don't add it here... add it to docs/apps.rst and wasp/boards/simulator/main.py instead (simulator has no RAM limit so it becomes a showcase for all the upstream applications).

I am looking at better ways to enable/disable applications but for now we are putting optional apps into the app library.

wasp/apps/musicplayer.py
```diff
@@ -0,0 +1,208 @@
...
+        self._play_icon = icons.play
+        self._pause_icon = icons.pause
+        self._play_state = False
+        self._icon_state = self._play_icon
+        self._gb_play = '{"t":"music", "n":"play"} '
+        self._gb_pause = '{"t":"music", "n":"pause"} '
+        self._gb_next = '{"t":"music", "n":"next"} '
+        self._gb_prev = '{"t":"music", "n":"previous"} '
+        self._gb_vup = '{"t":"music", "n":"volumeup"} '
+        self._gb_vdo = '{"t":"music", "n":"volumedown"} '
```
daniel-thompson, line 42: This is gadgetbridge protocol... it needs to be part of gadgetbridge.py

daniel-thompson (changes requested):
There's some "tough love" in this review ;-)

Let me know what you can and/or want to tackle from the review comments. Overall I'm keen to get this merged since a music player app is already on the TODO list for the 0.4 release!

Carglglz:
Sorry for the blank links, it was the auto formatting on save, I wasn't aware 🤦 .

I did update this according to the requested changes, just mention that it's more of a "concept minimal music app" rather than a definitive version, so feel free to make any future changes/improvements.

Regarding to RAM usage I would like to see some stats too. In the colmi P8 default build I remember seeing around 20-25 kB of free memory. Right now in my custom build I'm at around 10 kB (due to another custom app and fonts), so I don't think enabling this by default for the 0.4 shouldn't be too problematic. 👍

daniel-thompson:
Applied. Thanks.

PR #90, opened 2020-10-26
Adjust pulse vibration

cr0mbly:
### Summary

While playing around with the notifications on the Colmi P8 I've noticed that I often miss messages as the default vibration pulse is rather quick and weak. Playing around with the vibration class I've found the default set to .225 seconds to be more of reliable prompt. 

This is a personal taste and feel free to close off and ignore but I thought it was worth raising for discussions if anyone else thinks it might be worth a bump.

Signed-off-by: Aidan Houlihan aidandhoulihan@gmail.com

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,7 @@ cscope.out
 docs/build
 attic/
 wasp/boards/*/watch.py
+
+# IDEs
+.vscode
+.idea
--- a/wasp/drivers/vibrator.py
+++ b/wasp/drivers/vibrator.py
@@ -26,7 +26,7 @@ class Vibrator(object):
         self.period = 16000
         self.active_low = active_low
 
-    def pulse(self, duty=25, ms=40):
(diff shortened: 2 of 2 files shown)


Carglglz:
> the default vibration pulse is rather quick and weak

I agree with that but I was thinking of adding a configuration setting like the brightness, so one can adjust it at will.

Carglglz:
Test this: https://github.com/Carglglz/wasp-os/tree/settings-notify
I set three levels:
- **Silent**: no vibration
- **Mid**: Default light vibration (40 ms)
- **High**: Notable vibration (225 ms)

cr0mbly:
:+1: @Carglglz Tested and working a treat on my P8. Can you make a `Very high` setting with a 450 ms run time. I'm still finding that in a daily use 225 is still occasionally missed.

Carglglz:
Done in #93

PR #93, opened 2020-10-29
Notification Settings

Carglglz:
This PR adds a Notification level setting with four levels:
- Silent 
- Mid (40 ms pulse)
- High (225 ms pulse)
- Very High (450 ms pulse)

The Notification level can be adjusted with a slider widget.
This supersede #90

### DIFF
--- a/wasp/apps/settings.py
+++ b/wasp/apps/settings.py
@@ -19,6 +19,7 @@ class SettingsApp():
 
     def __init__(self):
         self._slider = wasp.widgets.Slider(3, 10, 90)
+        self._nfy_slider = wasp.widgets.Slider(4, 10, 90)
         self._settings = ['Brightness', 'Notification Level']
         self._sett_index = 0
         self._current_setting = self._settings[0]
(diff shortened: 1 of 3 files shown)


cr0mbly:
keen to have this in master, am using it now.

daniel-thompson:
Generally this looks like a good thing but I'm don't want to apply the "Very High" level.

It looks to me like the very high level is working around a bad PWM setting in the vibrator driver (P8 and PineTime motors are very different... currently the code treats them the same which makes the P8 vibrations far too weak).

Carglglz:
> It looks to me like the very high level is working around a bad PWM setting in the vibrator driver (P8 and PineTime motors are very different... currently the code treats them the same which makes the P8 vibrations far too weak).

I wasn't aware of the difference, so I changed it back to three levels (0, 40, 80) for PineTime and (0,225,450) for P8, although it may be better to address this at driver level?

daniel-thompson:
Merged! Thanks.

PR #95, opened 2020-10-31
Add remove all notifications dialog

cr0mbly:
### Summary

One of the annoying things with being spammed with notifications is to remove them you need to one by one view them on the watch to clear them all. I've added an extension to the `NotificationApp` that allows a user on a right swipe to clear them all:

clear notifications

I've also added the handy widget `ConfirmationView` that should be able to be used in the future in other places to confirm actions around the OS.

Have tested on my P8 watch but be interested to get @daniel-thompson testing on his pinetime as well to confirm it works/looks fine there as well.

Signed-off-by: Aidan Houlihan <aidandhoulihan@gmail.com>


wasp/widgets.py
```diff
@@ -285,3 +285,80 @@ def touch(self, event):
...
             v = self._steps - 1
         self.value = v
+
+
+_message_string_x_coord = const(0)
+_message_string_y_coord = const(0)
+_yes_button_x_coord = const(20)
+_yes_button_y_coord = const(100)
+_no_button_x_coord = const(120)
+_no_button_y_coord = const(100)
```
cr0mbly, line 295: these coords will only really look right as long as we continue to only support 240x240 displays we may have to make this more dynamic in the future. But I just copied the patterns I've seen in other places and they're pretty hard coded

wasp/apps/pager.py
```diff
@@ -82,19 +82,53 @@ class NotificationApp(PagerApp):
 
     def __init__(self):
         super().__init__('')
+        self.confirmation_view = wasp.widgets.ConfirmationView()
+        self.is_in_confirmation_view = False
```
daniel-thompson, line 86: Can self.is_in_confirmation_view be self.confirmation_view.active instead.

wasp/apps/pager.py
```diff
@@ -82,19 +82,53 @@ class NotificationApp(PagerApp):
...
 
+    def swipe(self, event):
+        mute = wasp.watch.display.mute
+        if event[0] == wasp.EventType.UP:
+            if self._page >= self._numpages:
+                wasp.system.navigate(wasp.EventType.BACK)
+            self._page += 1
+        elif event[0] == wasp.EventType.LEFT:
+            self.confirmation_view.draw('clear notifications?')
+            self.is_in_confirmation_view = True
```
daniel-thompson, line 111: I'd rather see this triggered with an additional swipe down (instead of the tactile pulse).

wasp/apps/pager.py
```diff
@@ -82,19 +82,53 @@ class NotificationApp(PagerApp):
...
+        else:
+            if self._page <= 0:
+                wasp.watch.vibrator.pulse()
+                return
+            self._page -= 1
+
+        if not self.is_in_confirmation_view:
```
daniel-thompson, line 118: If you move this test into the draw() method then you can use the super-class swipe method to manage up/down scriolling. Then this method ends up more like:

if event[0] == wasp.EventType.DOWN and self._page == 0:
    self.cv.active = True
    mute = wasp.watch.display.mute
    mute(True)
    self._draw()
    mute(False)
else:
    super().swipe(event)

wasp/apps/pager.py
```diff
@@ -82,19 +82,53 @@ class NotificationApp(PagerApp):
...
         del notes[id]
         title = note['title'] if 'title' in note else 'Untitled'
         body = note['body'] if 'body' in note else ''
         self._msg = '{}\n\n{}'.format(title, body)
 
+        wasp.system.request_event(
+            wasp.EventMask.TOUCH
+            | wasp.EventMask.SWIPE_LEFTRIGHT
+        )
         super().foreground()
```
daniel-thompson, line 102: To code to deactivate the confirmation view in the background() method seems to be missing (it means physical button and display timeout will not be handled correctly).

wasp/apps/pager.py
```diff
@@ -82,19 +82,53 @@ class NotificationApp(PagerApp):
...
+
+        if not self.is_in_confirmation_view:
+            mute(True)
+            self._draw()
+            mute(False)
+
+    def touch(self, event):
+        if self.is_in_confirmation_view:
+            result = self.confirmation_view.touch(event)
+            if result == self.confirmation_view.TouchResult.YES:
```
daniel-thompson, line 126: Do we really need the extra results type. If looks like True, False, None should be sufficient.

wasp/widgets.py
```diff
@@ -285,3 +285,80 @@ def touch(self, event):
...
+
+class ConfirmationView:
+    "Confirmation widget allowing user confirmation of a setting"
+
+    class TouchResult:
+        YES = 0
+        NO = 1
+        NO_ACTION = 2
```
daniel-thompson, line 303: Do we really need this special type? It looks like these are synonyms for True, False, None .

wasp/widgets.py
```diff
@@ -285,3 +285,80 @@ def touch(self, event):
...
+
+    def __init__(self):
+        self.drawable = wasp.watch.drawable
```
daniel-thompson, line 306: Don't cache the drawable in the instance. It does not have the same effect as caching within a single function.

wasp/widgets.py
```diff
@@ -285,3 +285,80 @@ def touch(self, event):
...
+
+    def touch(self, event):
+        x_coord = event[1]
+        y_coord = event[2]
+        is_yes_button_press = (
+            x_coord > self.yes_button_bounds[0][0]
+            and y_coord > self.yes_button_bounds[0][1]
+            and x_coord < self.yes_button_bounds[1][0]
+            and y_coord < self.yes_button_bounds[1][1]
+        )
```
daniel-thompson, line 350: This seems a little over precise (making the user hit a precise hit box doesn't really help us figure out their intent... it just means that have to press again). For this type of dialog it might be better to divide the screen info quarters and position the button to bottom left implies yes and bottom right implies no.

daniel-thompson (changes requested):
I'm on the fence on whether this is a good idea or not.

It looks rather like it is a workaround to a missing feature in GadgetBridge. In other words a better design is that if you "Clear All" on your phone then the watch notifications are also cleaned up without you having to trigger a special dialog at all.

However I also had some comments about implementation. In particular the down/down seems better than down/left to trigger a special dialog. However I'd still prefer it not to be needed at all.

cr0mbly:
> It looks rather like it is a workaround to a missing feature in GadgetBridge. In other words a better design is that if you "Clear All" on your phone then the watch notifications are also cleaned up without you having to trigger a special dialog at all.

This is a pretty common feature on a lot of smartwatches as someone usually will have the notifications loaded onto the watch while they're working and not notice they're being added, then check their phone and view them but they'll still be present on the watch.

I noticed it a lot with the low battery notification that was coming through. I had to view them all 1 by 1 in order to clear them from the watch after it had already been set to charge on the phone. This will be a similar case for other notifications

wasp/widgets.py
cr0mbly: Have changed to your suggestion, I felt that it was better to be explicit over the what the three states represent rather then just returning bools/nulls. The other advantage is there's a distinct value difference then between the types, where False and None can be lumped togeather as falsy

wasp/widgets.py
cr0mbly: :+1: cool thanks have changed

wasp/apps/pager.py
cr0mbly: Have refactored this to be a bit cleared, thanks for the suggestion

wasp/widgets.py
cr0mbly: To me this makes sense as the intent is conveyed via the message we sent to the view, the value is what's captured from the resulting `touch()` method. I think I'd find it weird as a user if I touch a out of bounds area of the screen and it still selected yes/no on the result. I know accidentally tap parts of my screen all the time.

daniel-thompson: It's not really a matter of accepting out-of-bounds touches. More about designing the UI to afford a larger hit box. As a simple example consider the launcher app which is designed without hard borders so that accepting presses in simple quardrents matches users expectation.

wasp/apps/pager.py
daniel-thompson: > This is a pretty common feature on a lot of smartwatches as someone usually
> will have the notifications loaded onto the watch while they're working and not
> notice they're being added, then check their phone and view them but they'll
> still be present on the watch.

Indeed... but the key point is that this is a bug in the companion app (it failed to clear the notifications from the watch when it was dismissed from the phone) and that fixing that is a better design.

Anyhow... I am happy to merge this since with the additional upward swipe it does fit fairly neatly into the UI. However that doesn't stop it being a workaround for a problem that should not exist ;-) .

One final thing though... the app isn't handling display timeout correctly. If you get the dialog up and then let the watch time out and turn off then the dialog needs to be automatically deactivated (otherwise the next time the user triggers the notification page it goes straight to the confirmation dialog).

daniel-thompson:
[Sorry if you get this twice. I posted it to the wrong bit of github the first time round ;-) ]

> This is a pretty common feature on a lot of smartwatches as someone usually
> will have the notifications loaded onto the watch while they're working and not
> notice they're being added, then check their phone and view them but they'll
> still be present on the watch.

Indeed... but the key point is that this is a bug in the companion app (it failed to clear the notifications from the watch when it was dismissed from the phone) and that fixing that is a better design.

Anyhow... I am happy to merge this since with the additional upward swipe it does fit fairly neatly into the UI. However that doesn't stop it being a workaround for a problem that should not exist ;-) .

One final thing though... the app isn't handling display timeout correctly. If you get the dialog up and then let the watch time out and turn off then the dialog needs to be automatically deactivated (otherwise the next time the user triggers the notification page it goes straight to the confirmation dialog).

cr0mbly:
> One final thing though... the app isn't handling display timeout correctly. If you get the dialog up and then let the watch time out and turn off then the dialog needs to be automatically deactivated (otherwise the next time the user triggers the notification page it goes straight to the confirmation dialog).

Ahh good point! have changed it now to deallocate the widget on termination:

deallocation


Think we're all good to merge now, thank you for your help :)

wasp/wasp.py
```diff
@@ -25,7 +25,7 @@
 from apps.flashlight import FlashlightApp
 from apps.heart import HeartApp
 from apps.launcher import LauncherApp
-from apps.pager import PagerApp, CrashApp, NotificationApp
```
cr0mbly, line 28: Unused import Replaced by `NotificationApp`

daniel-thompson:
Merged. Thanks!

PR #97, opened 2020-11-06
Added GitHub Actions support

daniel-thompson:
This is Panagiotis Vasilopoulos's PR rebased and converted into a PR so that I can experiment with the workflow.


panos:
Completely forgot about it, I'm glad that this is out of the way at least.

(The way we agreed to do this was way too complicated and the workflow in its current state may fail arbitrarily)

PR #101, opened 2020-11-13
Add wget to build dependencies

arteeh:
Wget needs to be installed to run `make softdevice`. Most distros have it installed by default, but it's better to have it in the list to be sure.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -19,7 +19,7 @@ following commands:
 .. code-block:: sh
 
     sudo apt install \
-      git build-essential libsdl2-2.0.0 python3-click python3-numpy \
+      wget git build-essential libsdl2-2.0.0 python3-click python3-numpy \
       python3-pexpect python3-pil python3-pip python3-serial unzip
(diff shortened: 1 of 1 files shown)


daniel-thompson:
Thanks for the PR. Could you add the Signed-off-by: (check out the [contributors guide](https://wasp-os.readthedocs.io/en/latest/contributing.html) for more details)

arteeh:
I can't get it to work. I enter `git commit --amend --signoff` like the guide explains, then `git push https://github.com/arteeh/wasp-os HEAD`. This rejects because it's behind remote, so I run `git pull --signoff` and `git push https://github.com/arteeh/wasp-os HEAD` again. The test runs and fails. Not sure what I'm doing wrong.

daniel-thompson:
Hi Maarten

I don't want to put you through a lot of pain for such a small patch. I think the problem was probably not using --force when pushing after amending the description. The subsequent merges seem to confuse things... when I checked your branch out locally and rebased it then all looked OK and I merged it.

Thanks!

PR #102, opened 2020-11-17
Added a Connect Four game application for two players.

Kozova1:
This PR introduces a Connect Four game to wasp-os. It requires 2 players to play, so it probably shouldn't be included by default.


Kozova1:
The game only needs a proper icon now I think

Kozova1:
I think I messed up this PR

PR #105, opened 2020-11-21
Add Alarm app

Siroj42:
This add an application to set a vibration alarm. The application allows setting a time and also activating/deactivating the alarm completely. If an alarm is over, it is automatically set again for the next day.
Application:
alarm py-screenshot
Application when ringing:
alarm py-alarm_screenshot


wasp/apps/alarm.py
```diff
@@ -0,0 +1,155 @@
+# SPDX-License-Identifier: LGPL-3.0-or-later
+# Copyright (C) 2020 Daniel Thompson
+# Copyright (C) 2020 Joris Warmbier
+"""Alarm Application
+~~~~~~~~~~~~~~~~~~~~
+
+An application to set a vibration alarm. All settings can be accessed from the Watch UI
```
daniel-thompson, line 7: Please add a screenshot to the header comment. This gets things ready for the app to go into the [Application Library](https://wasp-os.readthedocs.io/en/latest/apps.html). You can use the 's' button in the simulator to generate the screenshot and check ouf the Game of Life app ([wasp/apps/gameoflife.py](https://github.com/daniel-thompson/wasp-os/blob/master/wasp/apps/gameoflife.py) ) for an example of how to link to a screenshot from the header text.

wasp/apps/alarm.py
```diff
@@ -0,0 +1,155 @@
...
+"""
+
+import wasp
+import icons
```
daniel-thompson, line 11: This is unused.

wasp/apps/alarm.py
```diff
@@ -0,0 +1,155 @@
...
+import time
+
+icon = (
```
daniel-thompson, line 14: Please keep the comment giving the original file name and size (and also add the original png file to the res/ directory to make future changes easier).

wasp/apps/alarm.py
```diff
@@ -0,0 +1,155 @@
...
+    def foreground(self):
+        """Activate the application."""
+        self._draw()
+        wasp.system.request_event(wasp.EventMask.TOUCH)
+        if self.ringing:
+            wasp.system.request_tick(1000)
+
+    def background(self):
+        """De-activate the application."""
+        wasp.system.cancel_alarm(time.mktime(self.current_alarm), self._alert)
```
daniel-thompson, line 61: I think it might be better to cancel the alarm from the foreground() method. Seems needlessly difficult to test the app if the callback can trigger when it is already running.

wasp/apps/alarm.py
```diff
@@ -0,0 +1,155 @@
...
+        wasp.system.switch(self)
+
+    def _set_current_alarm(self):
+        now = wasp.watch.rtc.get_localtime()
+        yyyy = now[0]
+        mm = now[1]
+        dd = now[2]
+        if self.hours < now[3] or (self.hours == now[3] and self.minutes <= now[4]):
+            dd += 1
+        self.current_alarm = (yyyy, mm, dd, self.hours, self.minutes, 0, 0, 0, 0)
```
daniel-thompson, line 155: It would be better to store the output of time.mktime in self.current_alarm (it will use less RAM when the app is in the background).

daniel-thompson (changes requested):
Thanks for the PR. There are review comments but they are pretty minor.

The only other thing is that, ideally, the spinners would be integrated into widgets.py but I won't block merging the app for that if you'd rather leave that for something else.

tht-jxny:
I will be very happy when this gets merged. What is a smartwatch without an alarm? Many thanks, @Siroj42 for your effort!

Izaic:
I too would really like to see this merged! Looks awesome!

daniel-thompson:
I've been thinking of adding some text to the contributors guide to recommend against adding "please merge this" types of comments to pull requests (although there's nothing wrong with the PR author asking if there's anything more they need to do to). To be clear it's not that anyone is unwelcome! Quite the opposite. Simply that there are more effective forms of advocacy. In particular if you want to encourage something to be merged it is more helpful to provide details of how you tested it and on what device.

daniel-thompson:
PS With the 0.3 release out of the door I am now clearing the backlog of pull requests so, as it happens, in this case this PR will get looked at shortly.

daniel-thompson:
Merge! Thanks.

I also added the app to the library: https://wasp-os.readthedocs.io/en/latest/apps.html#alarm-application

PR #111, opened 2020-11-27
Added Connect 4 game app.

Kozova1:
This is an implementation of the classic connect 4 game.
It is meant to be played by two players, alternating turns.


Kozova1:
I messed up a few older PRs, this one should be the final one hopefully

Kozova1:
This crashes every time you place a piece on the actual device, so it'll need more testing.

daniel-thompson:
Hi @Kozova1 

I discovered that on Friday when I tried to merge it but I was trying to come up with a helpful summary of why and haven't had time yet.

Basically there are two missing features from micropython that will make recoding the win tests a little frustrating. Firstly micropython does not support pre-canned slices and nor does it support slicing of arrays with a step count arr[42:63:8].

Anyhow I did a few experiments with the Unix micropython port (which has all the available micropython language features enabled) I think you can rewrite this to replace the slices with tuples that can be passed to the range() function. You can then write a normal range loop to check the win conditions.

Something a bit like the following (I tested individual lines but not the whole function ;-) ):

    def is_win(arr, slice_):
        total = 0
        max_ = 0
        for i in range(*slice_):
            total += arr[i]
            max_ = max(max_, arr[i])
        return max_ and total == 4*max_
 
    a_diag = (37, 63, 8)
    if is_win(arr, a_diag):
       yay()

Hope that helps...


Daniel.

daniel-thompson:
Ah yes... I also realized one other thing... "Connect 4" is a trademark which means we will have to call it something different in wasp-os. I think "4-in-a-row" (if it fits in the launcher) or "Drop Four" are both good alternatives.

Kozova1:
> Hi @Kozova1 
> 
> I discovered that on Friday when I tried to merge it but I was trying to come up with a helpful summary of why and haven't had time yet.
> 
> Basically there are two missing features from micropython that will make recoding the win tests a little frustrating. Firstly micropython does not support pre-canned slices and nor does it support slicing of arrays with a step count arr[42:63:8].
> 
> Anyhow I did a few experiments with the Unix micropython port (which has all the available micropython language features enabled) I think you can rewrite this to replace the slices with tuples that can be passed to the range() function. You can then write a normal range loop to check the win conditions.
> 
> Something a bit like the following (I tested individual lines but not the whole function ;-) ):
> 
>     def is_win(arr, slice_):
>         total = 0
>         max_ = 0
>         for i in range(*slice_):
>             total += arr[i]
>             max_ = max(max_, arr[i])
>         return max_ and total == 4*max_
>  
>     a_diag = (37, 63, 8)
>     if is_win(arr, a_diag):
>        yay()
> 
> Hope that helps...
> 
> 
> Daniel.

Thanks! I'll look into implementing that

Kozova1:
Closed for now, if anyone else wants to they can fork my connect4 branch and continue development.

PR #115, opened 2020-12-05
apps: settings: Add Scroll Indicator

Kozova1:
This PR adds a ScrollIndicator widget to the settings app,
to help users realize they can scroll there.


wasp/apps/settings.py
```diff
@@ -80,4 +81,10 @@ def _update(self):
             else:
                 say = "Silent"
             self._nfy_slider.update()
+        settings_length = len(self._settings)
+        norm_idx = self._sett_index % settings_length
+        self._scroll_indicator.up = norm_idx > 0 and settings_length > 1
+        self._scroll_indicator.down = ( norm_idx < settings_length - 1
+                                        and settings_length > 1)
```
daniel-thompson, line 88: This is be over-engineered. The settings app is a loop of pages rather than a list. Thus you can always scroll both up and down making it better just to lose this code.

daniel-thompson (changes requested):
Thanks for this. It's a good change (and I was really pleased to see the screenshot updated as well).

Just one thing...

wasp/apps/settings.py
Kozova1: sure, will update in a few minutes.

Kozova1: done!

daniel-thompson:
Applied. Thanks!

PR #116, opened 2020-12-05
widgets: NotificationBar: Changed bluetooth icon to be blue.

Kozova1:
This might be somewhat subjective, so I can understand if this is not accepted.


daniel-thompson:
wasp-os current visual style is roughly speaking a mixture of ideas from GNOME 3 and Android... both of which use white/grey for the BT icon.

That means I definitely won't take this code as it is, however...

I might conisder a patch that introduces some kind of simple theming engine. It needs to be very respectful of RAM usage though... the raw storage needs to be a bytearray (wasp.sysem._theme) but there can be a helper function wasp.system.theme() to lookup a string ('clock', 'bt', ...) and return an RGB565 value.

PS I'm going to be adding something to the self-test to help us observe memory costs... I may still have to bounce the theme idea if we can't get things small enough (or find somewhere else for savings). It's starting to get a little cramped.

Kozova1:
I'm gonna look at implementing a theming engine and will submit a PR if I manage it

Kozova1:
see #117

Kozova1:
> I might conisder a patch that introduces some kind of simple theming engine. It needs to be very respectful of RAM usage though... the raw storage needs to be a bytearray (wasp.sysem._theme) but there can be a helper function wasp.system.theme() to lookup a string ('clock', 'bt', ...) and return an RGB565 value.

If we don't have to support hot reloading, `bytes` seems to be about twice as small as `bytearray`, at the cost of not being mutable.

PR #117, opened 2020-12-05
Added basic theming engine.

Kozova1:
This theming engine uses an array of 16bit values,
stored as `wasp.system._theme`.
It has a default value, which should not change anything about the way this looks currently.

The theme can be set via `wasp.system.set_theme`,
but this should **ONLY** be used in `main.py`.

To theme this, one has to use `tools/themer.py` (use flag -h for complete explanation)
to generate a bytestring that's added in main.py (see diff).

The bytestring is then converted to an array, and loaded into `wasp.system._theme`.
Theme values can be looked up by apps by using `wasp.system.theme("theme-key")`.
Theme keys appear in the function body of `wasp.system.theme()`.

I've took the liberty of converting existing apps to use this method,
and it seems to work well.

See the Alarm app for examples for apps, and the `ConfirmationView` widget for a widget theming example.

To add a theme value, it has to be added to the default theme class (`tools/themer.py`) and to the `DefaultTheme.serialize()` method.  
After that, modify `wasp.system.theme()` to include the wanted key, **in the order determined by the** `DefaultTheme.serialize()` **method**.

A test theme is provided in `tools/test_theme.py`


Kozova1:
This *definitely* has to appear in the documentation, just not sure where to put it.

Kozova1:
Currently, the theming engine supports setting colors for:  
- BLE icon in status bar
- Scroll indicator up arrow
- Scroll indicator down arrow
- battery charging icon
- battery empty icon
- battery full icon
- status bar clock
- confirmation view yes button
- confirmation view no button
- notification icon in status bar
- watchface left hours and left minutes digits
- watchface right hours and right minutes digits
- watchface colon

Izaic:
Wow! Very cool!

Kozova1:
Also tested this on my pinetime, seems to work perfectly! (at least I didn't notice any problems)

wasp/apps/clock.py
```diff
@@ -30,6 +30,10 @@ class ClockApp():
     """
     NAME = 'Clock'
     ICON = icons.clock
+    _colon_color = 0xb5b6
+    _left_color = 0xbdb6
+    _right_color = 0xffff
+    _read_theme = False
```
daniel-thompson, line 36: Caching theme colours in class variables is a waste of RAM (there is a trade off but RAM is usually more precious than CPU time). These should be looked up in the draw method.

wasp/boards/simulator/main.py
```diff
@@ -22,4 +22,5 @@
         'artist': 'Dreams of Bamboo',
     })
 
+wasp.system.set_theme(b'\xef{\xef{\xef{\xef{\x00\xf8\xe0\x07<\xe7iJiJ\xef{\xb6\xb5\xb6\xbd\xff\xff')
```
daniel-thompson, line 25: This is test code... it should not be commited.

wasp/main.py
```diff
@@ -3,4 +3,6 @@
 
 import wasp
 from gadgetbridge import *
+
+wasp.system.set_theme(b'\xef{\xef{\xef{\xef{\x00\xf8\xe0\x07<\xe7iJiJ\xef{\xb6\xb5\xb6\xbd\xff\xff')
```
daniel-thompson, line 7: Also test code... should not be committed.

wasp/wasp.py
```diff
@@ -504,4 +507,27 @@ def schedule(self, enable=True):
...
         self._scheduling = enable
 
+
+    def set_theme(self, new_theme: bytes):
+        self._theme = array.array('H', new_theme)
+    def theme(self, theme_part: str) -> int:
+        """Returns the relevant part of theme. For more see ../tools/themer.py"""
+        theme_parts = ("ble",
+                       "scroll-indicator-up",
+                       "scroll-indicator-down",
```
daniel-thompson, line 517: Why would up/down ever need to be be different colours?

wasp/wasp.py
```diff
@@ -504,4 +507,27 @@ def schedule(self, enable=True):
...
+                       "battery-charging",
+                       "battery-empty",
+                       "battery-full",
```
daniel-thompson, line 520: Similarly, why would charging and full be different colours.

wasp/wasp.py
```diff
@@ -504,4 +507,27 @@ def schedule(self, enable=True):
...
+                       "status-clock",
+                       "confirm-yes",
+                       "confirm-no",
```
daniel-thompson, line 523: I don't think confirm-yes/no should be candidates for themeing.

wasp/wasp.py
daniel-thompson: I'm not entirely sure battery-empty should be themeable either to be honest (the idea is that as the battery depletes the colour changes... and below a certain level the battery level gets too small to see so we colourize the frame with the colour the battery would be if we could see it).

wasp/wasp.py
```diff
@@ -504,4 +507,27 @@ def schedule(self, enable=True):
...
+                       "notify-icon",
+                       "watch-colon",
+                       "watch-left",
+                       "watch-right")
```
daniel-thompson, line 527: Watch specific theme entries are not a great idea (we wouldn't want to give that to all applications). Should be probably be based on accent level: "accent-hi", "accent-lo", "accent-mid" instead so other apps can adopt if they wish.

daniel-thompson: Also one of the more obvious things to theme is the slider widget and that is not here ;-)

wasp/wasp.py
```diff
@@ -118,6 +119,8 @@ def __init__(self):
         self.musicstate = {}
         self.musicinfo = {}
 
+        self._theme = array.array('H', b'\xef{\xef{\xef{\xef{\x00\xf8\xe0\x07<\xe7iJiJ\xef{\xb6\xb5\xb6\xbd\xff\xff')
+
```
daniel-thompson, line 123: To be honest I was expecting direct lookup from the bytestring (e.g. encode the colour as one byte and reuse the _clut8_rgb565() functions (I think these are currently best documented in tools/rle_encode.py tool).

For spot colours I would expect the 256 colour palette to be sufficient.

daniel-thompson (changes requested):
Overall its a great start.

There are some detailed comment below but overall:

* Don't cache the theme colours in class variables... it is a waste of RAM
* We need to review more carefully exactly what should be themeable
* Can we adopt the clut8 palette used by 2-bit RLE format so we can look up colours directly from a bytestring

wasp/boards/simulator/main.py
Kozova1: this is not quite test code - it's the default theme, which uses the current wasp-os colors

wasp/main.py
Kozova1: same as in `wasp/boards/simulator/main.py`

wasp/wasp.py
Kozova1: I disagree - a red no button and a green yes button is an example I could come up with

wasp/wasp.py
Kozova1: I didn't really understand this, could you elaborate a bit?

wasp/apps/clock.py
Kozova1: sure, should I change it for the widgets as well?

wasp/wasp.py
Kozova1: no problem, will rename "battery-charging" to "battery-charging-full" and will remove battery-full

wasp/wasp.py
Kozova1: idk, in the next revision they won't be.

wasp/main.py
Kozova1: In the next commit it's removed.

wasp/boards/simulator/main.py
Kozova1: removed in next commit

wasp/wasp.py
Kozova1: Slider has a `color` parameter, the only thing that can be changed maybe is the default color.

wasp/apps/clock.py
Kozova1: changed for widgets too.

wasp/wasp.py
Kozova1: I changed the theme storage to a `bytes` object (aka a `b'string'`).
This change will conserve a tiny bit of ram and will make it easier to set the theme.

wasp/apps/clock.py
Kozova1: only widget that was left caching is the Slider widget, and that's because it was already doing caching by itself.

It also does some computations with the color that I don't want to mess with, so it's the only widget atm that's still caching.

wasp/wasp.py
Kozova1: note that GNOME 3 also has colored buttons sometimes.

Kozova1: image
image

see images for examples ^

daniel-thompson: >> I don't think confirm-yes/no should be candidates for themeing.
>
> I disagree - a red no button and a green yes button is an example I could come up with

That's an argument for colouring the buttons! It doesn't have anything to do with theming. Note that the current YES/NO icons don't match the visual style of the rest of wasp-os so I'm already working on an update to the confirmation view that will change the artwork and add colour to the buttons.

However adding something to a colour palette theme system like the one you're working on implies that is makes sense to draw the UI component in a different colour and that doing so will make the user feel happier when they glance at their watch.

I simply don't think the colour of No/Yes buttons are good candidates for theming. There are only a narrow set of colour themes that make sense and users seldom see the dialog anyway. IMHO the best way to make the user(s) happy when they do see these buttons is to make them more attractive rather than the make the current ones themeable.

Kozova1: okay, will remove the confirmation dialog from the theming engine.

wasp/wasp.py
Kozova1: I'm not really sure what the benefits of using the CLUT here is.
We'd only save a few bytes at the cost of losing a lot of color options

I also think using the CLUT would make it harder for users to theme the watch by themselves - there are plenty of rgb565 color choosers online, but I'm not aware of any one of them that  supports wasp-os's CLUT.

Kozova1: See attached `script` session:

https://hastebin.com/perahimete.rb

daniel-thompson: > I'm not really sure what the benefits of using the CLUT here is.
> We'd only save a few bytes at the cost of losing a lot of color options

Converting the bytestring into an array makes a copy of the underlying data. Since the bytestring can be stored in flash it means we are comparing against a RAM overhead very close to zero.

Whether the theme uses a 65536 colour palette or a 256 colour palette doesn't matter nearly as much ensuring that wasp.system._theme is not format converted before being stored in the system class. If it started in flash in must stay in flash.


> I also think using the CLUT would make it harder for users to theme
> the watch by themselves - there are plenty of rgb565 color choosers
> online, but I'm not aware of any one of them that supports wasp-os's CLUT.

The 2-bit RLE encoder contains a function to lookup the nearest CLUT entry from a given RGB values. That can be used in the themer.

Note also that online colour choosers are pretty much useless for PineTime... the colour response is totally different to that of a typical RGB monitor. This is why there is a colour chooser in the selftest application... it's more or less impossible to design (attractive) colour applications for pinetime without using the built-in colour chooser.

Izaic:
Sorry to butt in, but I've seen you mention ram a lot lately. So with this patch in place there wont be any more room for apps on the Pinetime? How dire is the ram situation?

Can one disable internal tools such as the color slider tool and whatnot to save ram and make more space for apps? I have a pack of sealed Pinetimes, so i could easily have one Pinetime for Waspos as my daily driver with the dev tools disabled, and one with apps disabled and all the dev tools enabled.

Kozova1:
> Sorry to butt in, but I've seen you mention ram a lot lately. So with this patch in place there wont be any more room for apps on the Pinetime? How dire is the ram situation?
> 
> Can one disable internal tools such as the color slider tool and whatnot to save ram and make more space for apps? I have a pack of sealed Pinetimes, so i could easily have one Pinetime for Waspos as my daily driver with the dev tools disabled (self-test), and one with apps disabled and all the dev tools enabled.

I didn't measure the size of this PR, but it should be tiny (on device space wise).

It's absolutely possible to remove the self test application, have a look at `wasp/wasp.py` and find the relevant class.

tools/test_theme.py
```diff
@@ -0,0 +1,16 @@
...
+    BLE_COLOR = 0xfb80
+    SCROLL_INDICATOR_COLOR = 0xf800
+    BATTERY_CHARGING_FULL_COLOR = 0x07ff
+    SMALL_CLOCK_COLOR = 0x599f
+    NOTIFICATION_COLOR = 0x8fe0
+    ACCENT_MID = 0xf800
+    ACCENT_LO = 0x001f
+    ACCENT_HI = 0x07e0
+    SLIDER_DEFAULT_COLOR = 0x7777
+    pass
```
daniel-thompson, line 16: No `pass` needed here.

tools/themer.py
```diff
@@ -0,0 +1,75 @@
...
+"""Compiles themes for wasp-os"""
+
+from argparse import ArgumentParser, RawTextHelpFormatter
+from importlib import import_module
+from typing import Tuple
+
+class DefaultTheme():
+    """This class represents the theme. Import this file and extend the Theme class, only changing the variables.
+Export the resulting class as 'Theme'.
+"""
```
daniel-thompson, line 11: Indentation should match the start of the docstring (there is magic in the docstring parsing to allow sensibble indentation). I think it should also be

~~~
    """A single summary sentance.

    More detailed description spanning
    multiple lines and indented to match the opening
    quote.
    """"
~~~

wasp/apps/clock.py
```diff
@@ -38,6 +38,7 @@ def foreground(self):
         tick callback every second.
         """
         wasp.system.bar.clock = False
+
```
daniel-thompson, line 41: Don't mix non-functional and functional changes in the same patch.

wasp/wasp.py
```diff
@@ -504,4 +505,29 @@ def schedule(self, enable=True):
 
         self._scheduling = enable
 
+    def set_theme(self, new_theme) -> bool:
+        """Sets the system theme.
```
daniel-thompson, line 509: Newline missing here.

wasp/widgets.py
```diff
@@ -99,6 +100,8 @@ def update(self):
         now = wasp.watch.rtc.get_localtime()
         on_screen = self.on_screen
 
+            
+
```
daniel-thompson, line 104: This looks like "litter" left from previous changes.

wasp/widgets.py
```diff
@@ -206,10 +211,13 @@ def draw(self):
     def update(self):
         """Update from scrolling indicator."""
         draw = watch.drawable
+        
         if self.up:
-            draw.rleblit(icons.up_arrow, pos=self._pos, fg=0x7bef)
+            draw.rleblit(icons.up_arrow, pos=self._pos,
+                         fg=wasp.system.theme('scroll-indicator'))
```
daniel-thompson, line 217: Ideally we grab this once and store it in a local (locals are a bit special in micropython because the go on a stack and don't have names... so they don't consume memory in the same was as other idenfiers).

wasp/widgets.py
```diff
@@ -221,12 +229,17 @@ def update(self):
...
         self.value = 0
         self._steps = steps
         self._stepsize = _SLIDER_TRACK / (steps-1)
         self._x = x
         self._y = y
+        # required here unfortunately,
+        # this widget has setup code related to color,
+        # so it has to be reinitialized once theme is read.
+        self._read_theme = color is not None
+        color = color or 0x39ff
```
daniel-thompson, line 242: Better just set self._color to color here (and handle color being None in the draw method).

wasp/widgets.py
```diff
@@ -246,6 +259,10 @@ def draw(self):
         color = self._color
         light = self._lowlight
 
+        if not self._read_theme:
+            self.__init__(self._steps, self._x, self._y, wasp.system.theme('slider-default'))
+            self._read_theme = True
+
```
daniel-thompson, line 265: self._lowlight also needs to be set None in `__init__()` since we can't calculate the lowlight colour until we've decided what the highlight colour is going to be.

wasp/widgets.py
```diff
@@ -326,12 +343,12 @@ def draw(self, message):
         wasp.watch.drawable.blit(
             icons.yes_button,
             _yes_button_x_coord,
-            _yes_button_y_coord,
+            _yes_button_y_coord
```
daniel-thompson, line 346: Again, no need to modify unrelated code.

daniel-thompson (changes requested):
Looking good!

All the remaining review comments (perhaps except the lowlight colour stuff) are pretty fussy. Would be great if you can fix these up although if this PR is still pending the next time I get some wasp-os time then I might just do that myself.

wasp/apps/clock.py
Kozova1: this seems to be an artifact from my testing, will make sure to remove those

wasp/widgets.py
Kozova1: This was left as an artifact of the confirmation dialog theming, will remove

Kozova1:
Please have a look and suggest changes for the slider code, it's a bit ugly at the moment, and I was considering moving the lowlight calculation to a function, to prevent repetition.

wasp/widgets.py
```diff
@@ -244,6 +250,18 @@ def draw(self):
         x = self._x
         y = self._y
         color = self._color
+        if self._color is None:
+            self._color = wasp.system.theme('slider-default')
+            color = self._color
+            # Automatically generate a lowlight color
+            if color < 0b10110_000000_00000:
```
daniel-thompson, line 257: I don't think you need to pull this out into a function. I think you can just `if self._lowlight is None:' here and set it to  None from `self.__init__()`

Kozova1: no need to check that - if self._color is none, lowlight has to be none too.

daniel-thompson: What about the other case then self._color is not None?

We can either calculate it in two places: `__init__` and `draw` or we can calculate it only in `draw`. If we calculate it only in `draw` then there is no need to pull it out into a function.

wasp/wasp.py
Kozova1: There's a problem here - battery-full and battery-charging *are* different colors on the watch right now

Kozova1: so either we merge both to green by default, or we turn both into gray by default, (which causes problems with the fill on my device), or we keep them separate

or we could just set one of them to be fixed

Kozova1: For now, I set it so that battery-full is always green, while battery-charging is themable.

wasp/widgets.py
Kozova1: we can set it so that it only generates it once in `draw`

Kozova1: please review the current implementation

daniel-thompson:
Applied. Thanks.

PR #120, opened 2020-12-07
tools: Update intelhex, hexmerge.py and nrfutil for Python 3.9

Siroj42:
There are problems in the intelhex and nrfutil submodules regarding compatibility with Python 3.9. I noticed them after trying to build wasp-os on Fedora 33. Luckily, they are already fixed in the respective repositories, so simply updating the submodules and copying the hexmerge.py from tools/intelhex/scripts did the job.


Izaic:
Bump...

@daniel-thompson Could we get this merged? This fix is pretty important.

tools/hexmerge.py
```diff
@@ -35,7 +35,7 @@
 
 """Merge content of several hex files into one file."""
 
-VERSION = '2.2'
+VERSION = '2.3.0'
```
daniel-thompson, line 38: Why bump the version number here?

daniel-thompson (changes requested):
I've had a similar fix on my own tree for a couple of weeks but I was waiting for nrfutil to get updated (so it didn't have to go the trouble of forking it). Glad to see that's OK now.

Other than the version number bump this looks good!

tools/hexmerge.py
daniel-thompson: Actually I realized why... but drop this change anyway! Better to replace tools/hexmerge.py with a symlink so that's what I've just done.

shulltronics:
Hey guys,

Could anyone outline how to make this fix manually before the pull request is merged? I'm having trouble compiling wasp-os, and I think it's related to this. Thanks!

Izaic:
> Hey guys,
> 
> Could anyone outline how to make this fix manually before the pull request is merged? I'm having trouble compiling wasp-os, and I think it's related to this. Thanks!

A quickie temporary solution would be to just build wasp-os from the branch of this patch. While it will be older, you can at least explore a bit until the issue is resolved, and then after that you'll be able to try out the new stuff such as the haptic alarm.

daniel-thompson:
Applied. Thanks.

Issue #118, opened 2020-12-06
Add Line drawing function to draw565 library


Kozova1:
I noticed that the drawing library apparently does not support lines.
Is it possible to add a line drawing function?

I tried hacking in a line function but couldn't figure out how to write directly to the screen.

Kozova1:
Ideally, it would have an API like this:

`watch.drawable.line(x1,y1,x2,y2, color)`

maybe even add a thickness option?

daniel-thompson:
No objections here. One of the Bressenham variations should work fine.

Kozova1:
If you can help me understand how raw writing to the display works (quick_write etc.) I'll get a PR for it.

daniel-thompson:
quick_write() (which is designed to stop you having to keep pointlessly toggling the chip select) won't do you any good here because you can only write a couple of bytes at a time. You'll need to use display.set_window() to select the "rectangle" you want to update (start of with a 1x1 rectangle) and then use display.write() to draw the pixel.

Note drawing angled lines directly into the video RAM will be very slow because there's pretty much no way to optimise the transfer of data.
It will be very slow...

Kozova1:
see pr #124

PR #124, opened 2020-12-11
drivers: draw565: Added line() drawing function.

Kozova1:
This is the API:
`drawable.line(x1, y1, x2, y2, color)`

~Note: we have to find a better way to draw single pixels,
currently it uses fill() calls which are very expensive.~

The function has optimizations for the case of vertical or horizontal lines.

Signed-off-by: Kozova1 <mug66kk@gmail.com>

### DIFF
--- a/wasp/draw565.py
+++ b/wasp/draw565.py
@@ -355,3 +355,52 @@ class Draw565(object):
             chunks.append(end)
 
         return chunks
+
+    def line(self, x0=0, y0=0, x1=0, y1=0, color=0xffff):
+        """Draw a line between points (x0;y0) and (x1;y1) with color.
+
+        Example:
+
+        .. code-block:: python
+
+            draw = wasp.watch.drawable
+            draw.line(0, 120, 240, 240, 0xf800)
+
+        :param x0: X coordinate of the start of the line, defaults to 0
+        :param x1: Y coordinate of the end of the line, defaults to 0
+        :param y0: Y coordinate of the start of the line, defaults to 0
+        :param y1: Y coordinate of the end of the line, defaults to 0
+        :param color: Color to draw line in, in RGB565 format, defaults to 0xffff
+        """
+
+        px = bytes(((color >> 8) & 0xFF, color & 0xFF))
+        write_data = self._display.write_data
+        set_window = self._display.set_window
+
+        dx =  abs(x1 - x0)
+        sx = 1 if x0 < x1 else -1
+        dy = -abs(y1 - y0)
+        sy = 1 if y0 < y1 else -1
+        err = dx + dy
+        if dx == 0 or dy == 0:
+            if x1 < x0 or y1 < y0:
+                x0, x1 = x1, x0
+                y0, y1 = y1, y0
+            w = 1 if dx == 0 else dx
+            h = 1 if dy == 0 else -dy
+            self.fill(color, x0, y0, w, h)
+            return
+        while True:
+            set_window(x0, y0, 1, 1)
+            write_data(px)
+            if x0 == x1 and y0 == y1:
+                break
+            e2 = 2 * err
+            if e2 >= dy:
+                err += dy
+                x0 += sx
+            if e2 <= dx:
+                err += dx;
+                y0 += sy;
+        
+
--- a/wasp/apps/testapp.py
+++ b/wasp/apps/testapp.py
@@ -23,7 +23,7 @@ class TestApp():
     ICON = icons.app
 
     def __init__(self):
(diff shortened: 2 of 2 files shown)


Kozova1:
I need some help with implementing a `pix` function (color a single pixel).

Kozova1:
Updated the code, it now uses `display.set_window()` in combination with `display.write_data()`

Kozova1:
closes #118

wasp/draw565.py
```diff
@@ -355,3 +355,47 @@ def line(n):
...
+        :param x0: X coordinate of the start of the line, defaults to 0
+        :param x1: Y coordinate of the end of the line, defaults to 0
+        :param y0: Y coordinate of the start of the line, defaults to 0
+        :param y1: Y coordinate of the end of the line, defaults to 0
+        :param color: Color to draw line in, in RGB565 format, defaults to 0xffff
+        """
+        def _pix(x, y, color):
+            self._display.set_window(x, y, 1, 1)
+            self._display.write_data(((color >> 8) & 0xFF, color & 0xFF))
+            # no need to set_window() again, just wastes time.
```
daniel-thompson, line 378: This function is only called from one place so you'd probably be better off caching the internals of this method in local variables and then expanding the actual calls to set_window and write_data below:

~~~
set_window = self._display.set_window
write_data = self._display.write_data
px = bytes((color >> 8) & 0xff, color & 0xff))
~~~

This will remove the memory allocation (and several dictionary lookups) out of the inner loop.

Avoiding dictionary lookups doesn't make *that* much difference to benchmarks but avoiding allocations can. Talking of benchmarks can you add a simple line drawing benchmark into the self tests... doesn't need to be sophisticated just draw a bunch of lines in 15 degree increments and measure how long it takes.

wasp/draw565.py
```diff
@@ -355,3 +355,47 @@ def line(n):
...
+                break
+            e2 = 2 * err
+            if e2 >= dy:
+                err += dy
+                x0 += sx
+            if e2 <= dx:
+                err += dx;
+                y0 += sy;
+        # Reset the window to normal state
+        self._display.set_window()
```
daniel-thompson, line 399: There is no "normal state" for the window... so no need to reset to it!

wasp/draw565.py
Kozova1: will add benchmarks later, implemented the other changes.

Kozova1: Added benchmark and fixed bug with the vertical and horizontal lines optimization

wasp/draw565.py
```diff
@@ -355,3 +355,52 @@ def line(n):
...
+            draw.line(0, 120, 240, 240, 0xf800)
+
+        :param x0: X coordinate of the start of the line, defaults to 0
+        :param x1: Y coordinate of the end of the line, defaults to 0
+        :param y0: Y coordinate of the start of the line, defaults to 0
+        :param y1: Y coordinate of the end of the line, defaults to 0
+        :param color: Color to draw line in, in RGB565 format, defaults to 0xffff
+        """
+
+        px = ((color >> 8) & 0xFF, color & 0xFF)
```
daniel-thompson, line 376: Why is the bytes missing? This should be: px = bytes(((color >> 8) & 0xFF, color & 0xFF)) 

Has this code been tested on a real device?

daniel-thompson (commented):
Looks like github won't let me do a line-by-line reaction to the commit message... but there's no need to have two commit messages and sign-offs.

Once you squash things together there is only one patch so you can just have one message to describe what it does!

wasp/draw565.py
Kozova1: well this is embarrassing O.O

I'll fix it and test the fix again later today

Kozova1:
Yeah, probably should've used fixup instead of squash. Will be fixed together with the other issue later today

wasp/draw565.py
daniel-thompson: No worries.

The simulator will never cover everything a real device does but, as it happens, in this case I have been able to modify it to allow it to pick up these kind of problems: https://github.com/daniel-thompson/wasp-os/commit/e3b2c7bf7d7a4cc7dc2cafc28f35e30f44260a21

Kozova1: resolved, thanks

Kozova1: waiting for CI for a build I can test on my pinetime, will update with results.

daniel-thompson:
Applied. Thanks!

Kozova1:
branch line can now be deleted, right?

daniel-thompson:
Yes, once code it merged you can do whatever you want with the old branches.

PR #127, opened 2020-12-14
Added Heartrate app theming.

Kozova1:
This commit adds the option to change the color of the bars in the
heartrate app.

### DIFF
--- a/tools/test_theme.py
+++ b/tools/test_theme.py
@@ -9,7 +9,9 @@ class Theme(DefaultTheme):
     BATTERY_COLOR = 0x07ff
     SMALL_CLOCK_COLOR = 0x599f
     NOTIFICATION_COLOR = 0x8fe0
-    ACCENT_MID = 0xf800
-    ACCENT_LO = 0x001f
-    ACCENT_HI = 0x07e0
-    SLIDER_DEFAULT_COLOR = 0x7777
+    BRIGHT = 0xffff
+    MID = 0xff00
+    UI = 0x39ff
+    SPOT1 = 0x00ff
+    SPOT2 = 0x44d0
+    CONTRAST = 7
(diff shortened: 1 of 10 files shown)


daniel-thompson:
Similar to the feedback related to the clock I don't like the idea of theming each application individually.

A theme can offer a set of shades and colours that work together well but they should be named based on the character of the spot colour (bright, dull, etc) not the application that uses them.

For example

daniel-thompson:
So... I've been doing a bit of work on the widgets over the holidays and ended up theming everything from slider-default together with some lightening code to generate lighter and darker shades.

On that basis I'm tempted to consolodate app theming down to (for now I think we will leave the status bar theming as it is):

* accent-hi
* accent-lo
* ui-default (this is what is currently called slider-default)
* ui-bright
* ui-contrast

ui-bright will default orange and be picked up the HRS app.
ui-constast is used to generate lighter/darker shades of ui-default and ui-bright and dictates how much different they will be from the normal case.

I also propose a generate accent-mid using the the lighten() code since it is only a tiny bit different to accent-lo anyway. However I'm not 100% sure about this so I will need to prototype it to check it leaves the default apps looking OK.

Any objections to picking up this work?

Kozova1:
No problem - just don't forget to generate a default theme (to prevent unwanted crashes) :)

daniel-thompson:
Sorry... I think I made an unusually grave typo.

I had meant to type: "Any objections to me picking up this work?" then it looks like I omitted the "me" which leaves that last sentance *very* ambiguous! Anyhow it looks like your reply assumed I would have a go at this even with the typo ;-) .

daniel-thompson:
Hi @Kozova1 

I've been a bit brutal so I've pushed here to see what you think.

Daniel.

daniel-thompson:
I didn't see any review comments so I've just pushed these changes to master because I want to shoot a video on the new apps and theming capabilities today!

Kozova1:
Excited to see it :)

PR #134, opened 2020-12-24
Implement pedometer tap to reset

BNolet:
**This commit relies on the PR https://github.com/daniel-thompson/bma42x-upy/pull/1**

* Implement touch action in pedometer app
* Impelement more efficient step counter reset

### DIFF
--- a/wasp/apps/steps.py
+++ b/wasp/apps/steps.py
@@ -78,6 +78,13 @@ class StepCounterApp():
         self._wake += 24 * 60 * 60
         wasp.system.set_alarm(self._wake, self._reset)
 
+    def touch(self, event):
+        draw = wasp.watch.drawable
+        draw.fill(0x0000,60,112,180,40)
+        watch.accel.steps = 0      
+        
+        self._update()  
+
     def tick(self, ticks):
         self._count += 686;
         self._update()
--- a/wasp/drivers/bma421.py
+++ b/wasp/drivers/bma421.py
@@ -48,6 +48,8 @@ class BMA421:
     def steps(self, value):
         if value != 0:
             raise ValueError()
-        # TODO: There is a more efficient way to reset the step counter
-        #       but I haven't looked it up yet!
-        self.reset()
+        # TODO once bma42x.py and the library it relies 
+        # on are modified to correctly implement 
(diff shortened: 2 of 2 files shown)


wasp/apps/steps.py
```diff
@@ -78,6 +78,13 @@ def _reset(self):
         self._wake += 24 * 60 * 60
         wasp.system.set_alarm(self._wake, self._reset)
 
+    def touch(self, event):
+        draw = wasp.watch.drawable
+        draw.fill(0x0000,60,112,180,40)
+        watch.accel.steps = 0      
+        
+        self._update()  
+
```
daniel-thompson, line 87: I'm not entirely convinced by introducing touch to reset. I suspect acidental touching will be much more common than people who actually need to reset the counter. Additionally I've never seen another device that implements a full reset of the daily step count (since it would stop you reaching your daily step count goal).

There is certainly scope for adding a "trip counter" to the pedometer to be used to track runs or other exercise but I don't think this should be implemented by resetting with the daily step count.

BNolet: Good point, and that's sort of what I was going for. Just being able to use the counter as a trip counter, but maybe having a counter independent of the master counter would be useful.

What if the tap function makes another counter appear and then a second tap removes that second counter?

The second counter then just calculates a difference between when you first tapped and the present count

daniel-thompson: I'm not sure if it "wants" to be part of the step counter app or not. The main reason to wonder is that a trip counter pairs really well with a timer (many people who want a trip counter to measnre N thousand steps also want a timer to tell them how long it took them).

BNolet: So do you think it would be best to make this a completely separate app? Cuz if this is the case, I don't think it would pair particularly well with the existing stopwatch app either

daniel-thompson: Agree. Currently I'm thinking along the lines of an Activity or Sports app that can track both.

daniel-thompson:
Initial sports app has been added. See https://wasp-os.readthedocs.io/en/latest/apps.html#sports-timer

Graphically it's not very nice... but it implements the use-case we were discussing.

PR #137, opened 2020-12-26
Created memory-saving calculator app

tht-jxny:
This is a calculator app that saves memory space because no image is needed for displaying the buttons.

TODO: What has to be fixed is the font size. The app looks way better with font size 36 but unfortunately. all symbols (such as +,-,(,)) needed for calculations are missing in the `sans36` file. This has to be added!


daniel-thompson:
Hi @tht-jxny 

Rather than send another round of review comments I ended up making the changes directly.

Interestingly four of my changes are fixing bugs in draw565 and then *removing* the workarounds for those bugs from the calculator app!

Anyhow rather than just make brutal changes I posted them here before merging them.

Any objections to merging at it is now?

tht-jxny:
Thanks for all the improvements. Quite neat alternatives that I have not thought about!
For me, everything is fine. But what if ":" appears in the output. I thought it has to be replaced with a "/"?

daniel-thompson:
Applied. Thanks.

BTW In my last reply I neglected to take the time to say "great job". I did make quite a number of changes but that was, in part, because it was an inspired starting point!

PR #139, opened 2020-12-29
Added TabsApp(), which adds tabs to apps.

Kozova1:
To use it, when creating the app class inherit from it. For example, see this gist: https://gist.github.com/Kozova1/3015083dbb779c8d5fbd217f51e01a28

`TabsApp()` constructor takes a tuple of tab names.
It exposes the `draw()` method, which should be called *BEFORE* drawing to the screen,
since it resets the screen.

It also exposes the `tabs` and `current_tab` properties:
- `tabs`: the tuple passed to the constructor
- `current_tab`: the index to the tab (in `tabs`) that's currently focused.

The app is themable by changing `accent-lo` and `accent-hi` theme values.
To navigate between tabs,
`super.swipe(event)` should be called with the event as passed into `self.swipe()`.
`super.swipe(event)` takes `LEFT` and `RIGHT` swipes only, but will not crash otherwise.

### DIFF
--- a/wasp/apps/tabs.py
+++ b/wasp/apps/tabs.py
@@ -1,0 +1,56 @@
+"""Tabbed applications
+~~~~~~~~~~~~~~~~~~~~~~
+
+The tabs are used to provide multiple screens in apps.
+"""
+
+import wasp
+
+class TabsApp():
+    NAME = 'Tabs'
+
+    def __init__(self, tabs=()):
+        self.tabs = tabs
+        self.current_tab = 0
+
+    def foreground(self):
+        """Activate the app"""
+        self._redraw()
+
+    def swipe(self, event):
+        """Handle switching tabs"""
+        mute = wasp.watch.display.mute
+
+        moved = False
+
+        if event[0] == wasp.EventType.LEFT:
+            moved = self.current_tab < len(self.tabs) - 1
+            self.current_tab = min(len(self.tabs) - 1, self.current_tab + 1)
+        elif event[0] == wasp.EventType.RIGHT:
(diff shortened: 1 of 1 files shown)


Kozova1:
Note: this was not tested on an actual device.

Kozova1:
I'm not 100% sure about the style, see this picture:
image
(That's with 3 tabs, "Hello", "World", "Tab")

daniel-thompson:
I have to admit I'm not totally sold on this UI concept.

Currently there are two common "tabbed" apps, settings and self-test. Both of these are currently open coded (using a similar technique) rather than using a widget or super-class. Noteably they also use vertical paging rather than horizontal tabs and, in both cases I think will not really work with tabs as shown here (becaue there is or will not be enough horizontal space for the labels).

Kozova1:
currently, it uses draw.wrap() to truncate the strings to that they never overlap.

However, with a massive amount of tabs, it will overlap (if the tab width becomes < 1 char width)

daniel-thompson:
I'm afraid that, for me, the fact remains that the only existing apps where the tabs concept could be applied cannot use it because there isn't enough space for the tab labels.

daniel-thompson:
Sorry the close was a misclick... although I still really don't like the concept. In particular seeing games using horizontal swipes is fine but most regular applications should avoid them (since an application using horizontal swipes cannot be installed to the quick ring).

Note that the music app does currently use horizontal swipes (for Fwd and Back) but there are plans to change that.

Kozova1:
I'll close the PR for now.

PR #143, opened 2021-01-05
Add the 2048 game application

miguelrochefort:
2048 is a popular sliding block puzzle game in which tiles are combined to make the number 2048.

It's one of the few games that are enjoyable to play on such a small form factor.

This started as a port of a TkInter implementation of the 2048 game. I implemented all of the TkInter APIs used by the game and it worked on wasp-os without any code change in the game. However, the performance was very poor and it consumed too much RAM. I have since reimplemented the whole game from scratch and managed to achieve acceptable performance, although more improvements could still be made.

Because names in Python can't start with numbers, I had some trouble naming things. The module is called "ttfe" (two-thousand-forty-eight), the class name is Play2048App, and the software.py entry is "Play 2048".

The game was tested on the simulator and on the Colmi P8.


daniel-thompson:
Very impressive. As you mention there are still a few opporunities to reduce the footprint and I don't think I can cope with ttfe (too hard to remember, even *with* autocomplete) but this looks good for merging.

A couple of small things that are helpful (though I will fix them up this time around).

When adding new app then be sure to add the screenshot to README.rst and an suitable automodule stanza in docs/apps.rst. Also please use the simulator to take the screenshot (press the 's' key). This ensures the screenshot is the same size as all the other screenshots and that is has the correct filename for easy updates in the future.

miguelrochefort:
I'm available right now. 

Is there anything else I can do regarding this PR, or can I assume that "though I will fix them up this time around" means you'll take care of it?

daniel-thompson:
Merged. Thanks.

PR #145, opened 2021-01-07
Fix app unregister function

ChristianRomberg:
instanceof is not defined. I think you meant isinstance

### DIFF
--- a/wasp/wasp.py
+++ b/wasp/wasp.py
@@ -172,7 +172,7 @@ class Manager():
 
     def unregister(self, cls):
         for app in self.launcher_ring:
-            if instanceof(app, cls):
+            if isinstance(app, cls):
                 self.launcher_ring.remove(app)
                 break
 


daniel-thompson:
Oops... did I check that function in? That was a mistake. It is not used and should probably be deleted.

ChristianRomberg:
I find it quite useful tbh.

I made my own sort of "hot swap" so I don't always have to restart the sim.

```
import apps.scoreview as hs_module
class_name = "ScoreviewApp"

def get_hs_class():
	return getattr(hs_module, class_name)

from importlib import reload, import_module

class MyHandler(FileSystemEventHandler):
	def on_modified(self, event):
		print("on_modified", event.src_path)
		print("Reloading module")
		wasp.system.unregister(get_hs_class())
		reload(hs_module)
		wasp.system.register(get_hs_class()())

print("Registering event handler")
event_handler = MyHandler()
observer = Observer()
observer.schedule(event_handler, path="wasp/apps/", recursive=False)
observer.start()

wasp.system.register(get_hs_class()())
```

daniel-thompson:
Fair enough, in that case the change is OK for me... but can you fix the DCO autotest failure (see https://wasp-os.readthedocs.io/en/latest/contributing.html#introduction ).

ChristianRomberg:
Sure thing. Sorry for making you wait.

daniel-thompson:
Merged. Thanks!

PR #146, opened 2021-01-09
Update clock.py

masterdelic:
now displays plus day name in date, ex: Sun 10 Jan 2021


masterdelic:
wtf... please see my clock.py

PR #147, opened 2021-01-09
Update clock.py

masterdelic:
now displays plus day name in date, ex: Sun 10 Jan 2021

### DIFF
--- a/wasp/apps/clock.py
+++ b/wasp/apps/clock.py
@@ -19,6 +19,7 @@ DIGITS = (
 )
 
 MONTH = 'JanFebMarAprMayJunJulAugSepOctNovDec'
+DAY = 'ThuFriSatSunMonTueWed'
 
 class ClockApp():
     """Simple digital clock application.
@@ -96,6 +97,8 @@ class ClockApp():
         # Format the month as text
         month = now[1] - 1
         month = MONTH[month*3:(month+1)*3]
+        day = now[7] - 7
+        day = DAY[day*3:(day+1)*3]
 
         # Draw the changeable parts of the watch face
         draw.blit(DIGITS[now[4]  % 10], 4*48, 80, fg=hi)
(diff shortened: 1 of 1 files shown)


masterdelic:
ok i don't know how to do it..

SEE MY clock.py

wasp/apps/clock.py
```diff
@@ -19,6 +19,7 @@
 )
 
 MONTH = 'JanFebMarAprMayJunJulAugSepOctNovDec'
+DAY = 'ThuFriSatSunMonTueWed'
```
daniel-thompson, line 22: This lookup table doesn't really make any sense. The field that gives the day of the week is numbered 0..6 where 0 is Monday.

wasp/apps/clock.py
```diff
@@ -96,14 +97,16 @@ def _draw(self, redraw=False):
         # Format the month as text
         month = now[1] - 1
         month = MONTH[month*3:(month+1)*3]
+        day = now[7] - 7
```
daniel-thompson, line 100: now[7] is not the day of the week (0..6), it is the day of the year (0..364 except on leap years). In other words this code will break on Thursday!

now[6] is the field that gives the day of the week.

daniel-thompson (commented):
The other thing I should admit at this point is that I don't personally like the clock with the day of the week added. I prefer the shape of the display without the extra text.

However let's get the code working right for you before we worry about that!

wasp/apps/clock.py
masterdelic: When I start from Mon... I had a wrong day name displayed. It was Thursday when I did it that's why I start from Tue.
I made changes to now in 
draw.blit(DIGITS[now[*] 
  and see that 
0 is year 
1 is month
2 is day
3 is hour
4 is min
5 not found 
6 is sec
7 and success.

daniel-thompson: I don't understand how you are breaking up the localtime (how can 5 not be found?).

For me:

    >>> watch.rtc.get_localtime()
    (2021, 1, 10, 20, 23, 39, 6, 10)

Where:

* 0 is year (2021)
* 1 month (1)
* 2 is day of the month (10)
* 3 is hour (20)
* 4 is min (23)
* 5 is sec (39)
* 6 is day of the week (6 -> Sunday)
* 7 is day of the year (10 since currently day of the year == day of the month)

wasp/apps/clock.py
masterdelic: Now 6 yes you are right.... but still brake

masterdelic:
You really don't have to do that for me, I may need more reading

Issue #131, opened 2020-12-23
Lift to wake support


Kozova1:
Currently, wasp-os doesn't seem to support lift-to-wake (or if it does, it does so very poorly).  
This feature definitely should be considered.

BNolet:
If lift-to-wake gets put in, I also suggest tap-to-wake. That's supported in [hypnos](https://github.com/endian-albin/pinetime-hypnos) for the pinetime so it might be worth looking there. I'll open a separate issue if this feature isn't parallel enough of a feature to implement

Kozova1:
might just be me, but on my pinetime hypnos doesn't work with that

BNolet:
The tap to wake feature actually doesn't work in hypnos? :P

Issue #100, opened 2020-11-13
Colmi P8 StepCounterApp accelerometer driver crash


Gozel:
Hi Daniel, thanks for this open source project! 
I tried to install the steps.py on the Colmi P8 and I get following output after it's uploaded to the watch:
~~~
Traceback (most recent call last):
  File "<stdin>", in <module>
  File "<stdin>", in __init__
  File "drivers/bma421.py", in reset
ValueError: invalid sensor
~~~
Can you advise?

SteveAmor:
Hi @Gozel check out #69 
It's likely that you have a newer P8 with a different accelerometer.

daniel-thompson:
Assuming you bought the new model P8 (with the magnetic charger) then this is because the sensor is not supported:
https://wasp-os.readthedocs.io/en/latest/install.html#colmi-p8

Gozel:
Hi Steve and Daniel, yes, that's it. Thanks! Apologies for having overlooked that. 
Question, will there be support for the new accelerometer?

daniel-thompson:
I don't think anybody is currently working on this, so don't expect anything in the short term. Ultimately we never know when an interested person will step forward to write something so, in the longer term, I just don't know.

Gozel:
Fair. 
Many thanks!

PR #149, opened 2021-01-12
PR Demo: P8 SC7A20 accelerometer

Carglglz:
This is not intended to be merged just yet. This is just for demo purpose.

This PR includes:
- SC7A20 MicroPython driver.
- "Wake on shake" feature.

After some google/forums research it turns out that this accelerometer is pretty similar to [ST LIS3DH](https://www.st.com/resource/en/datasheet/cd00274221.pdf). So using SC7A20 data sheet (ft. google translate), the LIS3DH data sheet and its [Application Note](https://www.st.com/resource/en/application_note/cd00290365-lis3dh-mems-digital-output-motion-sensor-ultralowpower-highperformance-3axis-nano-accelerometer-stmicroelectronics.pdf) I managed to put together a basic SC7A20 MicroPython driver.

There are still other features to be implemented (activity detection, free fall...), but for now I think it is good for testing. 
#131

About the step counter #100,  I guess this needs to be implemented in code since SC7A20 does not have an step counter feature built in as the BMA421. 👀


wasp/boards/p8/manifest.py
```diff
@@ -11,7 +11,7 @@
     (
         'boot.py',
         'draw565.py',
-        'drivers/bma421.py',
+        'drivers/sc7a20.py',
```
daniel-thompson, line 14: Just a quick comment but there are plenty of P8 boards with BMA421 sensors (in fact I wear one most of the time). You probably need to create a second board type (P8b) for the newer devices.

Carglglz: You are totally right 🤦

leofidus:
I'm running this on my watch right now. I had to modify the Makefile a bit to get the P8b version to build, but Shake-to-wake is much better than I would have expected. You can just flick your wrist to turn the watch on, which feels much more versatile and predictable than traditional lift-to-wake

Carglglz:
>  I had to modify the Makefile a bit to get the P8b version to build

I should update that in the commit too 😅 

>  but Shake-to-wake is much better than I would have expected

This is why it was the first feature I tested, also shake-to-wake sensitivity could be adjusted with a setting option, I may work on that in the future...

leofidus:
> also shake-to-wake sensitivity could be adjusted with a setting option

yes, the current setting is a bit draining on the battery

I played around a bit with this and created an App that just graphs accelerometer values. Some thoughts in no particular order:

- accel() and acceleration seem to be doing kind of the same thing, just different (different ways to read it, accounting for different settings)
- the option to change the units used by accel() seems kind of nice at first, but in the end I'm hard pressed to name a use case? It's just another setting every app would have to set in foreground(). It's probably easier to standardize on either m/s^2 or g (I guess that's the same conclusion you came to by the time you implemented self.acceleration)
- m/s^2 fells more correct, but g is imho actually more ergonomic (values between -2 and 2, at rest the vector (x,y,z) has unit length, etc
- `accel.acceleration` is a very nice interface to use
- it would be nice to also have an endpoint that keeps everything in integers, for efficiency. Something like a `acceleration_microG` that adjusts for range and power mode but keeps everything in integers, in addition to `acceleration`'s float version
- lift-to-wake should actually be pretty easy to implement. I might take a stab at that
- step detection seems harder; sampling at 8 Hz walking looks distinctive (similar to offset sine waves on all three channels), but running needs more samples to be decipherable. I guess that would be an ideal use case for the FIFO functionality of the accelerometer

Carglglz:
> I played around a bit with this and created an App that just graphs accelerometer values

Nice, this could be a stand alone App or part of Test App too. 

>  Some thoughts in no particular order:

Yes, this should be taken as a draft version on which to build, adding or dropping whatever feels necessary. 👍

I agree with the `accel / acceleration` thoughts, and if you had a look at the Application Note you probably saw that there is still room for improvement and interesting features to be implemented.

> step detection seems harder; sampling at 8 Hz walking looks distinctive (similar to offset sine waves on all three channels), but running needs more samples to be decipherable. I guess that would be an ideal use case for the FIFO functionality of the accelerometer

Yes I think is difficult to find a step detection solution that is both reliable and low power. (Maybe adding another setting option to enable/disable would be a nice approach. 🤔 )

leofidus:
> Nice, this could be a stand alone App or part of Test App too.

Yes, I should make a pull request once I've cleaned it up a bit. 

> Yes I think is difficult to find a step detection solution that is both reliable and low power. (Maybe adding another setting option to enable/disable would be a nice approach. 🤔 )

A rough idea based on watching pretty graphs of me walking:

Operating mode of 25Hz, enabled FIFO with interrupt on INT1 (so one FIFO overflow every 1.28 seconds). On every interrupt from the accelerometer the code would check INT1_SRC, and if the interrupt is from the FIFO then read all 6*32=192 bytes and just put it into some internal buffer until there are maybe 10 seconds of data. Then check if the entire buffer has roughly regular peaks followed by regular valleys on x and z channels. If yes, then the number of peaks is the number of steps taken, if not then no steps were taken. From the number of steps you could even derive if it's walking or running. This seems reasonably low power. It probably has a lot of false positives and can't count small numbers of steps, but that's not that different from many older step counters.

Carglglz:
> A rough idea based on watching pretty graphs of me walking:

>Operating mode of 25Hz, enabled FIFO with interrupt on INT1 (so one FIFO overflow every 1.28 seconds). On every interrupt from the accelerometer the code would check INT1_SRC, and if the interrupt is from the FIFO then read all 6*32=192 bytes and just put it into some internal buffer until there are maybe 10 seconds of data. Then check if the entire buffer has roughly regular peaks followed by regular valleys on x and z channels. If yes, then the number of peaks is the number of steps taken, if not then no steps were taken. From the number of steps you could even derive if it's walking or running. This seems reasonably low power. It probably has a lot of false positives and can't count small numbers of steps, but that's not that different from many older step counters.

This could be actually a good approach 🤔

Also I've just realised that the graph App with data logging to a file would be ideal for debugging the walking/running or other physical activities. Then send this file to the computer, test the signal processing part and implement the best solution. 👍

daniel-thompson:
> A rough idea based on watching pretty graphs of me walking

I'd recommend writing a (throwaway) python script that captures the data (at whatever sample rate your research to date suggests is good) and writes it out to the filesystem. Providing it doesn't print() anything you can set the script running from the REPL then go for a walk and collect a fairly large corpus of test data at roughly the same step rate. When you get back you can reconnect, ^C and get the data back to your development machine (it will take quite a while I imagine). You can do the same for moderate running, flat-out sprinting and "just sitting talking, watching TV, hacking, etc" (e.g. basically anything except stepping). It is useful to record which data comes from which activity since just grabbing data and trying to categorising by eyeball when you get back is tedious ;-) .

Once you have the data you can sit in front of a workstation running jupyter and develop an algorithm by testing it against all the data you have captured. The main reason to work like this isyou will probably find that there are outliers and they can help you tune the algorithm (e.g. reduce false steps, avoid missed steps, etc). This only really works if you have a large enough corpus of data and sensible predicted results for them.

If you are wondering, the above is more or less how ppg.py was written. The ppg code has different challenges to step counting (few dimensions but PPG wrist sensors have crazy bad signal quality) but I suspect the same approach to developing algorithm will work.

daniel-thompson:
BTW you can, of course, also add data logging to an app instead of as a script but only if you are entirely confident the screen redraws don't interfere with the sampling rate.

puuu:
Hello, I am also interested in getting a step counter/pedometer to work on the P8b watch. 

How about implementing a documented or already otherwise implemented algorithm into wasp-os?

I found the following in a short search: 

Analog Devices provide a documented algorithm. This probably should also work with non ADXL axxelerometers: [Full-Featured Pedometer Design Realized with 3-Axis Digital Accelerometer by Neil Zhao](https://www.analog.com/en/analog-dialogue/articles/pedometer-design-3-axis-digital-acceler.html)

Another option would be source code from an Android App (Privacy Friendly Pedometer):
https://github.com/SecUSo/privacy-friendly-pedometer/blob/master/app/src/main/java/org/secuso/privacyfriendlyactivitytracker/services/AccelerometerStepDetectorService.java


There are probably more possibilities. IMHO these are already good candidates.

devmarxx:
Congratulations on integrating the SC7A20 driver for the p8b! 

I got everything working today on my p8b and I'm very pleased. It was a bit fiddly to remove the Bosch driver, but once I got that, the step counter app came up and the "wake on shake" started working :-)  

Wake on shake is way to sensitive, but that is nitpicking. It is great to have all sensors working on the p8b. 

Respect.

Chris

cgoates:
Hi, sorry if this is the wrong place to ask this, but could those who got this PR running share the modifications they made to the Makefile in order to do so?  I'm trying to hack it together on my own, but I'm missing something and can't get the compile to finish.

didboy:
> Congratulations on integrating the SC7A20 driver for the p8b!
> 
> I got everything working today on my p8b and I'm very pleased. It was a bit fiddly to remove the Bosch driver, but once I got that, the step counter app came up and the "wake on shake" started working :-)
> 
> Wake on shake is way to sensitive, but that is nitpicking. It is great to have all sensors working on the p8b.
> 
> Respect.
> 
> Chris

Hi Chris!  I've got the P8B version and I can't make the accelerometer work.  It seems that you reached to correct the code and enable it. Could you share the code modified?
Thanks!

devmarxx:
I did not continue working on this device and I have no documentation on what I did to make it work.

The Bosch driver I referred to is the "bma421", but I have no clue what I did with that back then. 

Sorry that I can be of more help.

Success ...

Chris

didboy:
> I did not continue working on this device and I have no documentation on what I did to make it work.
> 
> The Bosch driver I referred to is the "bma421", but I have no clue what I did with that back then.
> 
> Sorry that I can be of more help.
> 
> Success ...
> 
> Chris

Hi Chris,
Many thanks for your quick reply.   I continue finding out how to do it.  If I have any progress I'll share with all of you.
Regards,
Dídac

PR #150, opened 2021-01-13
Disallow swipes while confirmation prompt is displayed

rgegriff:
Signed-off-by: George Griffin <george@menomnom.com>


daniel-thompson:
Merged. Thanks for the fix.

PR #151, opened 2021-01-17
Feature/rgegriff/track step history

rgegriff:


rgegriff:
I took a crack at adding a step history to the StepCounterApp. If you have any previous day's steps in memory (up to a full week's worth) you can swipe down to cycle through them. It needs some clean up, but I am opening a PR now in case anyone wants to comment.

Since it stores the histories in ram, it will obviously reset if you reset your device. @daniel-thompson what are your thoughts around persisting settings/configs/datalogs/etc. in flash? I would think limiting writes to flash would be desirable, but I think the trade off of sparingly writing to flash could be worth it.

daniel-thompson:
Writing to flash is fine, especially if it is only 4 bytes per day.

In fact when writing 4 bytes per day the simplest way to avoid flash wear is *not* to have code to clean the log (no need to store dates for something periodic, we can calculate those from the filesize). It only writes at 1.5k per year so it would take decades before the file size is a problem ;-) .

rgegriff:
> (no need to store dates for something periodic, we can calculate those from the filesize)

I was thinking about that. My main concern there was the bug around the _reset() function where it won't get called if you are looking at the app. I suppose fixing that rather than storing the date in another 4 bytes might be the better approach.

I'll implementing this next time I am hacking on this.

daniel-thompson:
Hi @rgegriff 

After a few calculations I decided to get a bit more ambitious with the logging. I've just pushed some code to add a step counting service to wasp-os. It will log the step count in six minute intervals (e.g. 240 times a day to make graphing easy) which is still fairly cheap in terms of storage since the data rate is 480 bytes per day.

I'm not sure what the filesystem overhead is but I think we should be able to log for about 2 years without running out of storage... although at this date rate I probably will have to add some delete code too.

Anyhow, if you want to try it out then grab the latest sources!

> I'll try implementing this next time I am hacking on this.

I think that the above means there's no need to clean up this PR at this point. I hope that's OK.

rgegriff:
Nice! This is kind of the direction I wanted to inch toward before I got distracted with work stuff! ill check it out!

PR #153, opened 2021-01-20
st7789: Fix incorrect variable name in ST7789_SPIs. __init__ docstring.

tworaz:
The data signal pin name is dc, not cs.

Signed-off-by: Piotr Tworek <tworaz@tworaz.net>


daniel-thompson:
Applied. Thanks!

(and sorry for the delay)

PR #154, opened 2021-01-21
Added App to Graph Accelerometer Values

leofidus:
This is an app to show accelerometer values based on @Carglglz's driver in #149.

It looks like this (except that the title now displays properly): 

In that picture you can see me walking in the middle of the graph, followed by lifting the watch and me fiddling with my phone to take this picture. The graph wraps around, so the beginning of the photographed graph is me taking the picture.

Since the BMA421 driver doesn't support getting the actual values and I don't have a pinetime (yet), this app only does something meaningful on a newer P8 with #149 merged. It does fail gracefully though, displaying a message indicating the problem.

This PR is still missing a screenshot and a logo, but I don't think it's ready to merge anyways until #149 has a stable API and is merged. Until then treat it more as a proof-of-concept of what's possible.

Signed-off-by: Jannis Froese <jannisfroese@mailbox.org>


daniel-thompson:
Interesting feature. However I'm not sure it makes sense for this to be a stand alone app, I think it would make more sense as part of the self test app.

daniel-thompson:
Actually, on reflection if you want to add data logging and make this into a standalone accelerometer test app then that's OK... but please rename to Test Accelerometer (or Test Accel if it doesn't fit on the screen) to make sure it doesn't over-excite anyone in the Software app.

PS if you adopt this approach please have a look at the failed autotests...

leofidus:
I'm aware of the failed autotest, that's just the missing screenshot I mentioned above.

I would be a bit hesitant to add it as a 16th screen to the self test. I am considering to renaming the app to "Sensors" or something along those lines and adding a couple of additional screens (for now temperature, later maybe vibration frequency derived from the accelerometer, useful if you lay the watch flat on a vibrating surface).

daniel-thompson:
As far as I'm concerned the self test can have any many screens as necessary. If it gets really unmanageable we can replace the vertical scrolling to select them with something else. The purpose of the test app is to avoid an explosion of test apps that clutter up other parts of the UI like the Software enable tool or the launcher.

However a sensor test app is probably OK.

PR #156, opened 2021-01-21
Initial Lift-To-Wake implementation

leofidus:
This implements #131 for devices with a compatible accelerometer (so currently only #149).

I have implemented it as a "driver", allowing any board to use it with the available accelerometer driver or replace it with a different implementation (for example because of hardware lift-to-wake support).

To test the code, merge it with #149, and in boards/p8b/watch.py.in add `from drivers.lift_to_wake import LiftToWake` around line 36 and `lift_to_wake = LiftToWake(accel, rtc)` around line 99. Then add `'drivers/lift_to_wake.py',` around line 22 of boards/p8b/manifest.py, recomile, upload and you are done.

Usually "lift-to-wake" requires you to lift your arm and hold a position with the watch tilted by about 15 degrees. Getting a reliable lift detection that also works while walking seems hard, but it turns out that just detecting the end state is actually enough. You just have to hold you arm vertically, then tilt it slightly towards you. This works even with your arm resting on a surface.

Signed-off-by: Jannis Froese <jannisfroese@mailbox.org>

### DIFF
--- a/wasp/drivers/lift_to_wake.py
+++ b/wasp/drivers/lift_to_wake.py
@@ -1,0 +1,34 @@
+# SPDX-License-Identifier: LGPL-3.0-or-later
+# Copyright (C) 2020 Jannis Froese
+
+"""A pseudo-device to handle lift-to-wake
+~~~~~~~~~~~~~~~~~~~~~~~~
+"""
+
+class LiftToWake(object):
+    """Lift-to-Wake handler
+
+    .. a lift-to-wake implementation that uses an existing accelerometer without dedicated lift-to-wake support
+
+    .. automethod:: __init__
+    """
+
+    def __init__(self, accelerometer, rtc):
+        """Create a lift-to-wake handler"""
+        self.accel = accelerometer
+        self.rtc = rtc
+        self._updateAt = rtc.get_uptime_ms()
+        self.wakeCount = 0
+
+    def update(self):
+        """update internal state, returns true if device was lifted"""
+        if self.rtc.get_uptime_ms() < self._updateAt:
+            return False
+        self._updateAt = self.rtc.get_uptime_ms() + 125
+
+        (x, y, z) = self.accel.acceleration
+        #print(y < 1 and y > -3 and x < -2 and z < -3 and z - x < -3 and z - x > -6, y < 1, y > -1, x < -2, z < -3, z-x < -3, z-x > -6, z-x, x, y, z)
+        if y < 1 and y > -3  and x < -2 and z < -3 and z - x < -3 and z - x > -6:
+            self.wakeCount += 1
+            return True
+        return False
--- a/wasp/wasp.py
+++ b/wasp/wasp.py
@@ -140,6 +140,7 @@ class Manager():
         else:
             self._nfylevels = [0, 40, 80]
         self._nfylev_ms = self._nfylevels[self._notifylevel - 1]
+        self._lift_to_wake_enabled = 'lift_to_wake' in dir(watch)
         self._button = PinHandler(watch.button)
         self._charging = True
         self._scheduled = False
@@ -204,6 +205,19 @@ class Manager():
     def notify_duration(self):
         """Cached copy of the current vibrator pulse duration in milliseconds"""
         return self._nfylev_ms
+
+    @property
+    def lift_to_wake_supported(self):
+        return 'lift_to_wake' in dir(watch)
+
+    @property
+    def lift_to_wake_enabled(self):
+        return self._lift_to_wake_enabled
+    
+    @lift_to_wake_enabled.setter
+    def lift_to_wake_enabled(self, value):
+        self._lift_to_wake_enabled = value and self.lift_to_wake_supported
+    
 
     def switch(self, app):
         """Switch to the requested application.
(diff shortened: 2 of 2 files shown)


daniel-thompson:
Cool to see this sort of thing being worked on but I'm afraid I'm not sure about the way this is integrated into wasp-os.

Firstly what is the benefit of implementing algorithms as a pseudo driver? Basically seeing a gesture recognition algorithm special cases in the Manager class doesn't seem right to me. IMHO it should be library code (more similar in structure to ppg.py than a driver) that is called by accelerometer drivers for raw (e.g. not smart) sensors like the one in P8b.

leofidus:
I don't have a strong opinion about whether the functionality should be in a pseudo driver, in the accelerometer driver or in a library used by the accelerometer driver. Moving it from a pseudo driver to a driver library just moves two lines in the board setup to four lines in the accelerometer driver. I don't think a pseudo driver is unreasonable or unprecedented (the rtc driver is also just a wrapper around another driver), but I'm happy to move the code in a library.

I don't really get what you mean with the special cases in the Manager? Are you talking about the setup in the board (which a library would delegate to the accelerometer driver instead of the board) or the code in wasp.py (which wouldn't change in either case, the decision whether to keep the screen awake isn't a driver concern but a system manager concern, and failing gracefully if features are unimplemented is also clearly a system manager concern).

daniel-thompson:
Let me gently suggest that assuming exising wasp-os driver code is well organised is a mistake. The current driver code was hacked together at speed and, IIRC, *before* the current scheduling model was implemented. The plan was (and has always been) to tidy these things up when a second implementation comes along. In other words a second touchscreen driver will mean revisiting the touch screen event code and having a second application already caused us to tidy up the interfaces between the app and the manage (and that the first person to want a display that is not 240x240 will need to do quite a bit of plumbing first). Anyhow that means when adding a second accelerometer we must look at how to "do it right" (or at least do it  better).

So the main reasons I object to a pseudo driver to recognise the lift gesture is that it doesn't work for smart sensors such as BMA423 and that it, even for dumb sensors, it doesn't work well when the accelerometer driver adds interrupt support and we can improve the gesture recognition algorithm. IMHO the accelerometer driver should report a lift event to the system.

What I said about the system manager is also bound up on this. There should be no need for it to fail gracefully if the accelerometer (and it's driver) cannot detect the lift gesture since, if this case the acccelerometer won't report the event.

leofidus:
So to summarize you would prefer a model where the (accelerometer) driver does everything without help from the system, if it needs to run periodically does that itself (via interrupts from the hardware it manages, the system manager's alarm feature or one of the unused timers or RTCs), handles the necessary settings itself (similar to how the settings app currently directly interacts with the RTC driver), and only interacts with the system manager to tell it to do something (e.g. calling `wake()` or `wake(reason='lift')`)?

That's certainly a different distribution of responsibilities than I would have guessed from reading the existing codebase, but I agree that it is much better for smart hardware that can generate more useful interrupts, while not being notably worse for other cases (the only real disadvantage is that some lift-to-wake code would run even on an already awake system, at the very least enough to determine that the screen is already awake).

>So the main reasons I object to a pseudo driver to recognise the lift gesture is that it doesn't work for smart sensors such as BMA423

the board could just do

```
accel_driver = BMA423(i2c)
self.accel = accel_driver
self.lift_to_wake = accel_driver
```

as long as the driver APIs aren't overlapping in incompatible ways.

> and that it, even for dumb sensors, it doesn't work well when the accelerometer driver adds interrupt support

I agree, I don't anticipate the FIFO interrupt to be useful simply because lift-to-wake has to happen sub-second to be intuitive (500ms already feels too slow) while activity detection has no such requirement and thus would likely configure a slower FIFO. But I could imagine that orientation detection interrupts are useful on some dumb sensors.

daniel-thompson:
> So to summarize you would prefer a model where the (accelerometer) driver does everything without help from the system, if it needs to run periodically does that itself (via interrupts from the hardware it manages, the system manager's alarm feature or one of the unused timers or RTCs), handles the necessary settings itself (similar to how the settings app currently directly interacts with the RTC driver), and only interacts with the system manager to tell it to do something (e.g. calling `wake()` or `wake(reason='lift')`)?

Not quite like the Settings app since there's no apps involved in lift-to-wake and the system manager certainly involved to some degree. To be honest eventually there might need to be a shared event queue but its tricky to get right from a memory allocation point of view so giving the accelerometer a polled get_event() method is enough for now (like buttons and touch screen, etc).

BTW the general timers are not much use to any event generating driver on nRF52... they stop ticking when the device is sleeping.

> > and that it, even for dumb sensors, it doesn't work well when the accelerometer driver adds interrupt support
> 
> I agree, I don't anticipate the FIFO interrupt to be useful simply because lift-to-wake has to happen sub-second to be intuitive (500ms already feels too slow) while activity detection has no such requirement and thus would likely configure a slower FIFO. But I could imagine that orientation detection interrupts are useful on some dumb sensors.

I don't follow this. Doesn't the P8b sensor allow you to set the FIFO interrupt threshold to an interval suitable for gesture detection?

thiswillbeyourgithub:
I can't wait to see progress on that front :) any news?

PR #161, opened 2021-01-31
drivers: draw565: Added bounding_box() to check physical size of string

Kozova1:
Example:

```py
width, height = wasp.watch.drawable.bounding_box("Hello")
```

Warning: This was only tested in the simulator, but it hopefully should not break anything on the watch.

### DIFF
--- a/wasp/draw565.py
+++ b/wasp/draw565.py
@@ -330,6 +330,14 @@ class Draw565(object):
         if width:
             self.fill(bg, x, y, rightpad, h)
 
+    def bounding_box(self, s):
+        """Return the bounding box of a string.
+
+        :param s: A string
+        :returns: Tuple of (width, height)
+        """
+        return _bounding_box(s, self._font)
+
     def wrap(self, s, width):
         """Chunk a string so it can rendered within a specified width.
 


Kozova1:
This https://github.com/daniel-thompson/wasp-os/pull/161/checks?check_run_id=1802794751 check looks like a false negative to me.

wasp/draw565.py
```diff
@@ -329,6 +329,15 @@ def string(self, s, x, y, width=None, right=False):
         if width:
             self.fill(bg, x, y, rightpad, h)
 
+    def char_size(self, char):
+        """Get the size of a specific character in the current font.
+
+        :param char: Character to get the size of.
+        :returns     Tuple with  (width, height) of character.
+        """
```
daniel-thompson, line 337: The auto build is failing for this pull request and is reporting problems with this docstring. The missing : after returns is likely to cause this.

wasp/draw565.py
```diff
@@ -329,6 +329,15 @@ def string(self, s, x, y, width=None, right=False):
...
+        (_, h, w) = self._font.get_ch(char)
+        return (w, h)
```
daniel-thompson, line 339: Height will always be constant so it would be better to implement this as separate get_char_width() and get_char_height() methods since in almost all cases the h value will be discarded (therefore allocating a tuple to hold it should be avoided).

daniel-thompson:
Actually thinking about this I guess the other question is why have an interface for a single character in the first place? It might be better to calculate the length of a string?

wasp/draw565.py
Kozova1: good point - changes incoming.

Kozova1:
That's a good point - now we have str_width(string) and char_height(), both of which are pretty self explanatory.

daniel-thompson:
I've just been looking at this PR and it struck me that this is basically just a reimplementation of the existing _bounding_box() function (I'm a bit more relaxed about the tuple when it is not allocated per character).

Perhaps this can just collapse to be:

```
def bounding_box(s):
    return _bounding_box(self._font, s)
```

daniel-thompson:
Sorry to have gone hither and thither on this one. Anyhow, it is now merged. Thanks!

PR #163, opened 2021-02-02
drivers: draw565: Added arc() function to draw an arc of a circle.

Kozova1:
The function draws a part of the circle, using the specified color,
between the specified angles.
See docstring for better explanation.

The function is *very* slow, so it shouldn't be used too much.

Warning: only tested on the simulator.


Kozova1:
Important note: It currently does not support setting a stroke width, so all arcs are 1px wide.

Kozova1:
Currently, there's a bug where if you try to draw let's say 350 degrees (sweep) with start=50 it will only draw between 50 < a < 360.

Kozova1:
@daniel-thompson this was not yet tested on a physical device, only in the emulator.

(PR scope has changed quite a bit here, whoops!)

wasp/draw565.py
```diff
@@ -518,3 +518,43 @@ def darken(self, color, step=1):
...
+        When precision is higher, the function is *much* slower, but also draws nicer lines.
+        Control points define the curve, and should be a list of tuples of (x, y) coordinates.
+
+        Example:
+
+        .. code-block:: python
+
+            draw.bezier(0xf800, 100, [(0, 0), (1, 1), (2, 0)])
+
+        :param color:           A RGB565 color in which the curve will be drawn.
```
brvier, line 535: Put color at the end of parameters like other func ?

Kozova1: Color is at the start here somewhat intentionally - the list of points can get quite long, imo it makes a bit more sense to break that convention here.

daniel-thompson:
Are there any apps planned that need this feature?

Kozova1:
I myself don't have anything planned, but it could be useful for drawing general curves, and although bezier curves can't represent a circle perfectly, the approximation can get quite good - so this could remove the need for a circle / arc drawing functions.

PR #168, opened 2021-02-25
Add ability and setting to wake screen when a notification is received

althypothesis:
- Added the ability to wake the watch screen when a notification is received
- Added settings to the settings app to allow vibration and screen wake to be enabled or disabled by the user from the watch itself

I loaded this onto my own PineTime devkit hardware I use daily and it has been working for the last 24 hours without an issue. I left the default setting as "vibrate only" since that is what it always was.

This is my first _ever_ pull request and I do not know python. I did my best to follow existing conventions. If I did anything wrong, please let me know and I'll be happy to fix it. I've been using open source for over a decade and I am excited to finally be able to give back!

Signed-off-by: zack+github@nullh.net

### DIFF
--- a/wasp/apps/settings.py
+++ b/wasp/apps/settings.py
@@ -37,7 +37,9 @@ class SettingsApp():
         self._dd = wasp.widgets.Spinner(20, 60, 1, 31, 1)
         self._mm = wasp.widgets.Spinner(90, 60, 1, 12, 1)
         self._yy = wasp.widgets.Spinner(160, 60, 20, 60, 2)
-        self._settings = ['Brightness', 'Notification Level', 'Time', 'Date']
+        self._von = wasp.widgets.Checkbox(3, 40, 'Vibrate')
+        self._won = wasp.widgets.Checkbox(3, 80, 'Wake Screen')
+        self._settings = ['Brightness', 'Notification Level', 'Time', 'Date', 'Notifications']
         self._sett_index = 0
         self._current_setting = self._settings[0]
 
@@ -68,6 +70,12 @@ class SettingsApp():
                 now[1] = self._mm.value
                 now[2] = self._dd.value
                 wasp.watch.rtc.set_localtime(now)
+        elif self._current_setting == 'Notifications':
+            if self._von.touch(event):
+                wasp.system.vibe_on_notif = self._von.state
+            elif self._won.touch(event):
+                wasp.system.wake_on_notif = self._won.state
+            print(wasp.system.vibe_on_notif, wasp.system.wake_on_notif)
         self._update()
 
     def swipe(self, event):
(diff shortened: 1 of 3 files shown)


wasp/apps/settings.py
```diff
@@ -37,7 +37,9 @@ def __init__(self):
         self._dd = wasp.widgets.Spinner(20, 60, 1, 31, 1)
         self._mm = wasp.widgets.Spinner(90, 60, 1, 12, 1)
         self._yy = wasp.widgets.Spinner(160, 60, 20, 60, 2)
-        self._settings = ['Brightness', 'Notification Level', 'Time', 'Date']
+        self._von = wasp.widgets.Checkbox(3, 40, 'Vibrate')
```
daniel-thompson, line 40: I'm not sure that this checkbox is needed? If you want to disable Notifications then set them to Silent with the slider.

wasp/apps/settings.py
```diff
@@ -115,6 +123,11 @@ def _draw(self):
             self._dd.draw()
             draw.set_font(fonts.sans24)
             draw.string('DD    MM    YY',0,180, width=240)
+        elif self._current_setting == 'Notifications':
+            self._von.state = wasp.system.vibe_on_notif
+            self._won.state = wasp.system.wake_on_notif
+            self._von.draw()
+            self._won.draw()
```
daniel-thompson, line 130: It might be a good idea to add the new checkbox to the existing page (with the slider) and update the title.

wasp/wasp.py
```diff
@@ -205,6 +207,24 @@ def notify_duration(self):
...
+
+    @property
+    def wake_on_notif(self):
+        """Wake the screen when a notification is received bool"""
+        return self._wake_on_notif
+
+    @wake_on_notif.setter
+    def wake_on_notif(self, value):
+        self._wake_on_notif = value
+
```
daniel-thompson, line 227: There's no need for these to use properties... just a regular member variable (without the leading underscore) is fine.

daniel-thompson (changes requested):
Interesting idea. Needs a couple of tweaks but I think it could be useful (at least until we get lift-to-wake working with BMA421).

wasp/wasp.py
daniel-thompson: > Interesting idea. Needs a couple of tweaks but I think it could be useful (at least until we get lift-to-wake working with BMA421).

PS Can you take a look at the DCO failure as well (the Signed-off-by: needs to be in the commit rather than the PR). I think clicking on the Details next to the failure will give some hints on how to do this in git.

brvier:
I made similar change on my own watch but i switch only if screen is off as i could be using calc or an other app.

brvier:
Yes sorry i notice the pull request later and forgot to close this issue.

Izaic:
Anything new on this?

PR #169, opened 2021-03-03
Add roundRect method and some pep8

brvier:
Added roundRect method and his test in TestApp, and some pep8 format at the same time.

Not tested on a real device, only sim (still waiting mine).

### DIFF
--- a/wasp/apps/steps.py
+++ b/wasp/apps/steps.py
@@ -47,7 +47,7 @@ class StepCounterApp():
 
     def __init__(self):
         watch.accel.reset()
-        self._count = 0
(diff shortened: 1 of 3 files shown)


Kozova1:
Wouldn't this be better as 2 separate PRs? One for roundRect, and one for pep8?

Also, doesn't PEP8 say to use snake_case instead of camelCase?

brvier:
Indeed, bad golang dev habbits :)
-- 

Benoît HERVIER - https://rvier.fr


4 mars 2021 00:15:02 Kozova1 <notifications@github.com>:

PR #172, opened 2021-03-03
Dual clock (HH on top, MM on bottom)

brvier:
Add dual clock as an app.

image

We should probably add a setting to choose an app as clock.


daniel-thompson:
Please can you move the artwork used by the app into the app itself (rather than putting it in the shared code). I know this different to the default clock but the big images look like they have quite a large flash footprint so we may not be able to enable the clock by default.

Also can you take a look at the test results... there are a large number of failures.

daniel-thompson:
> Please can you move the artwork used by the app into the app itself (rather than putting it in the shared code). I know this different to the default clock but the big images look like they have quite a large flash footprint so we may not be able to enable the clock by default.

It is OK for the digits to be in fonts.clock_dual (since we can disable that if we need to) but integrating the icon into the app it probably a good plan.

brvier:
I don't understand, the unit test attend a DualApp.png image name ?

```
E       AssertionError: assert '.. figure:: res/DualApp.png' in 'Digital dual clock\n~~~~~~~~~~~~~~~~\n\nShows a time (as HH and MM vertically) together with a battery meter.\n\n.. figure:: res/DualClockApp.png\n    :width: 179\n'
```

daniel-thompson:
> I don't understand, the unit test attend a DualApp.png image name ?

Yes.

More specifically it expects the screenshot to be created using the simulator's snapshot feature (the 's' key) and the the file name may not be altered.

brvier:
I integrated icon in app, and made the screenshot with the simulator, then squash commits.

taitberlette:
That looks nice! Very clean and minimal.

> We should probably add a setting to choose an app as the clock.

Although having an option to change the clock face on the watch would be nice, I have an app for wasp-os which supports changing the clock face that works in the meantime. If you want to have a look at the app, it has a repository [here](https://github.com/taitberlette/wasp-os-companion).

daniel-thompson:
I think there's still some "pixel art" work that could be done to make this watch face more attractive.
It looks like the conversion to true 1-bit hasn't been hand edited for smoothness afterwards.

However there's no need to hold up the merge for that so... Applied. Thanks!

PR #178, opened 2021-03-04
New Dice and Coin app

FlameKat53:
please test on real hardware!


daniel-thompson:
Can you take a look at the testing failures above?

It may not be clear why the documentation is failing to build (and this is why the binary distribution test is failing). It is failing because you have added the application icon in icons and not updated the documentation version of icons.py. However, for non-core apps it is recommended to include the icon in the app itself so you can fix this by moving the icon out of icons.py!

FlameKat53:
> Can you take a look at the testing failures above?
> 
> It may not be clear why the documentation is failing to build (and this is why the binary distribution test is failing). It is failing because you have added the application icon in icons and not updated the documentation version of icons.py. However, for non-core apps it is recommended to include the icon in the app itself so you can fix this by moving the icon out of icons.py!

yes sir!

FlameKat53:
Failed to import ecdsa, cannot do signing
?

daniel-thompson:
Hey, if you are struggling getting the auto tests then it's easy to run them locally to make things quicker.

`make check` runs the simulator tests.

`make dist` runs the documentation and build tests.

FlameKat53:
Oh, thank you!

wasp/apps/fliproll.py
```diff
@@ -0,0 +1,118 @@
+ 
+# SPDX-License-Identifier: LGPL-3.0-or-later
+# Copyright (C) 2020 Daniel Thompson
+# Copyright (C) 2021 Evan Rodgers 'setLillie'
+
+"""Digital Coin Flip
+~~~~~~~~~~~~~~~~
```
taitberlette, line 7: I think this is one of the reasons why the checks are not successful. I had the same problem, and it wouldn't run successfully until the docstring (`~`) was the same length as the title. You can see the error in the build [here](https://github.com/daniel-thompson/wasp-os/pull/178/checks?check_run_id=2059068256#step:8:51).

wasp/apps/fliproll.py
```diff
@@ -0,0 +1,118 @@
...
+
+Shows a dice roll on screen tap
+
+.. figure:: res/dice_icon.png
```
taitberlette, line 11: Also, I just noticed this. In your commit there is no `res/dice_icon.png`. Did you mean `res/FlipRollApp.png`?

PR #184, opened 2021-03-16
tools: wasptool: can now check battery level

thefenriswolf:
This adds a battery check for wasptool, which I found quite handy at times.
Let me know if I need to change anything for this pr to be accepted.

Signed-off-by: thefenriswolf <stefan.rohrbacher97@gmail.com>

### DIFF
--- a/tools/wasptool
+++ b/tools/wasptool
@@ -189,6 +189,9 @@ def handle_exec(c, fname):
     print_log(log)
     log.close()
 
+def handle_battery_level(c):
+    print(f'Battery: {c.run_command("watch.battery.level()")}%')
+    
 def handle_memory_free(c):
     # wasp namespace to avoid having to import into the global space
     print(f'Before GC: {c.run_command("wasp.gc.mem_free()"):7}')
@@ -329,6 +332,8 @@ if __name__ == '__main__':
             description='Wasp-os command and control client')
     parser.add_argument('--as', dest='upload_as', default=None,
             help="Filename to use on the target (e.g. wasptool --upload docs/main/chrono.py --as main.py")
+    parser.add_argument('--battery', action='store_true',
+            help="Report remaining battery charge")
     parser.add_argument('--bootloader', action='store_true',
             help="Reboot into the bootloader mode for OTA update")
     parser.add_argument('--binary', action='store_true',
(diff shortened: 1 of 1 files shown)


wasp/fonts/__init__.py
```diff
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: LGPL-3.0-or-later
 # Copyright (C) 2020 Daniel Thompson
 
+import fonts.sans18 as sans18
```
Kozova1, line 4: I don't think this is related to the PR, is it?

wasp/fonts/sans18.py
```diff
@@ -0,0 +1,264 @@
+# Code generated by font_to_py.py.
```
Kozova1, line 1: Wouldn't this be better in a separate PR?

wasp/requirements.txt
```diff
@@ -1,6 +1,7 @@
 cbor
 click
 cryptography
+freetype-py
```
Kozova1, line 4: As I said above ^^

Kozova1 (changes requested):

wasp/fonts/sans18.py
thefenriswolf: You are absolutely correct, this was intended to be a separate PR.
I have no idea how I messed this up

Kozova1: If you'd run these commands it should fix it:

```sh
git rebase -i HEAD^^
```
And write "drop" next to the font commit.
After that, create a new branch and run this command:
```sh
git cherry-pick 919f8b86de5704a3749eb565a83b8c0b8d9d1d1d
```
And submit a separate PR for it.

Kozova1: Make sure that your new branch is based on **this** master, not the one you use - otherwise, your font PR will have both the font and the battery level changes.

Generally, it's easier to work with a master branch that's synced to this branch, and each time you want to add a feature branch from it and add your feature there.

Kozova1: Always glad to help :)

daniel-thompson:
Applied. Thanks.

PR #186, opened 2021-03-17
add 18pt font

thefenriswolf:
Signed-off-by: thefenriswolf <stefan.rohrbacher97@gmail.com>


daniel-thompson:
I've merged this for now... but we're very close to running out of flash so at some point there might be difficult decisions about keeping it ;-) .

Thanks!

PR #188, opened 2021-03-18
update requirements.txt

thefenriswolf:
ecdsa - for binary signing
freetype-py - required for generating fonts

### DIFF
--- a/wasp/requirements.txt
+++ b/wasp/requirements.txt
@@ -1,6 +1,8 @@
 cbor
 click
 cryptography
+ecdsa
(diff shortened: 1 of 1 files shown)


daniel-thompson:
I'm not sure if we want to merge these. We don't sign the binaries so we don't need ecdsa and whilst freetype-py is definitely useful for generating fonts nothing in the wasp-os build system ever generates fonts... I think it might be better to drop micropython-font-to-py from the tools/ directory instead and cover regenerating fonts in the documentation instead.

PR #195, opened 2021-03-23
fill_ppm  write ppm image format to screen

bmsleight:
Signed-off-by: Brendan M. Sleight <bms.git@barwap.com>

### DIFF
--- a/wasp/draw565.py
+++ b/wasp/draw565.py
@@ -518,3 +518,71 @@ class Draw565(object):
         b = bm - step if bm > step else 0
 
         return (r | g | b)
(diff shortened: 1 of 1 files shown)


daniel-thompson:
Has this been tested on a real device? If so what is the performance like? Using `fill()` for single (or quad) pixel access is normally very slow.

It would also be good to allow the caller to select an x, y coordinate for the decoding (e.g. give it a signature more like `_rle2bit()`.).

daniel-thompson:
No reply from submitter... converted to draft status.

PR #198, opened 2021-03-30
apps: Added weather application.

taitberlette:
Added a new weather application to view data provided by the Bangle.JS protocol [here](https://www.espruino.com/Gadgetbridge), so it should work with GadgetBridge, and I have tested it with my wasp-os companion. The UI is pretty basic, it doesn't have images yet, but it shows all the data needed. Let me know if I need to change/add anything to this pull request.


daniel-thompson:
Applied. Thanks.

PR #199, opened 2021-03-31
MemoryError exception now shows custom message instead of CrashApp.

taitberlette:
The MemoryError exception happens quite a bit for me, so I decided to make it a bit neater. Instead of showing the entire error message, the MemoryError shows the problem and suggests how to fix it.


daniel-thompson:
Applied. Thanks!

PR #210, opened 2021-05-17
alarm: Add a snooze button.

Liorst4:
There is a snooze button and a dismiss button.
Snooze adds 5 minutes to the alram.
Dismiss stops the alarm.
Pressing the home button is also acts as snooze.

### DIFF
--- a/wasp/apps/alarm.py
+++ b/wasp/apps/alarm.py
@@ -64,10 +64,14 @@ class AlarmApp():
         self.hours.value = 7
         self.ringing = False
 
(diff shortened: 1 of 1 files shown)


daniel-thompson:
Thanks for the PR. Please can you look at the reported test failures.

daniel-thompson:
I came within an hour or so of merging this but then #225 arrived, which includes Snooze (10 minute by default), multiple alarms and day of the week scheduling.

I'm afraid that means I (eventually) merged the more ambitious changes. I hope you like them and if you want to tune the Snooze behaviour from #225 then I'll certainly listen carefully.

PR #213, opened 2021-05-22
settings: Add option to choose the watch face.

Liorst4:
Removed watch apps from the Software app

### DIFF
--- a/wasp/apps/settings.py
+++ b/wasp/apps/settings.py
@@ -28,6 +28,16 @@ class SettingsApp():
     NAME = 'Settings'
     ICON = icons.settings
 
+    # TODO: dictionary / set instead of a tuple
+    # TODO: should be static?
+    FACES = (
+        ('clock', 'Clock'),
+        ('chrono', 'Chrono'),
+        ('dual_clock', 'Dual Clock'),
+        ('fibonacci_clock', 'Fibonacci Clock'),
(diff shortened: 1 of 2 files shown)


daniel-thompson (commented):
Thanks for the PR but I'm afraid I really don't like the UI concept for this, especially using checkboxes to implement radio buttons and having a hard limit on the number of watch faces.

I think we can do a much better job by using a custom app that shows previews of each watch face. I already have prototype code for this so I will see if that can be cleaned up.

daniel-thompson:
> I think we can do a much better job by using a custom app that shows previews of each watch face. I already have prototype code for this so I will see if that can be cleaned up.

Well... it took a little longer than I would have liked, but it cleaned up into this which I hope works well for you:
https://github.com/daniel-thompson/wasp-os/commit/1a4b678b2b815884cb712244d8a1b5e50a5cbd86

Liorst4:
> > I think we can do a much better job by using a custom app that shows previews of each watch face. I already have prototype code for this so I will see if that can be cleaned up.
> 
> Well... it took a little longer than I would have liked, but it cleaned up into this which I hope works well for you:
> [1a4b678](https://github.com/daniel-thompson/wasp-os/commit/1a4b678b2b815884cb712244d8a1b5e50a5cbd86)

Look great :+1:

PR #223, opened 2021-06-23
Fix escape sequences in wasptool

inducer:
I typically run with `PYTHONWARNINGS=default` (for other development work), and this PR helps avoid two recurring warnings when using `wasptool`:

```
/home/andreas/pack/wasp-os/./tools/wasptool:226: DeprecationWarning: invalid escape sequence \(
  c.expect('\(([0-9]+), ([0-9]+), ([0-9]+), ([0-9]+), ([0-9]+), ([0-9]+), ([0-9]+), ([0-9]+)\)')
/home/andreas/pack/wasp-os/./tools/wasptool:388: DeprecationWarning: invalid escape sequence \(
  console.expect('Connect.*\(([0-9A-F:]*)\)')
```

Also, thanks for all you do on waspos!

### DIFF
--- a/tools/wasptool
+++ b/tools/wasptool
@@ -223,7 +223,7 @@ def handle_rtc(c):
 
 def check_rtc(c):
     c.sendline('print(watch.rtc.get_localtime())')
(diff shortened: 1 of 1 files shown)


daniel-thompson:
Applied. Thanks.

PR #225, opened 2021-06-23
Advanced alarm app

beardeddude:
Features:
* Multiple alarms (up to 4)
* Day of the week support
* One time alarms
* Snooze

Changes to wasp-os for app support:
* Added + and - to the 28pt and 36pt fonts
* * I needed + on the 28pt font. I added - for consistency and added both to the 36pt for consistency as well
* Checkboxes now require a click on the body of the checkbox if there is no label
* * If there is a label the functionallity is the same
* Added a Toggle Button class that extends Button and stores a state like checkbox
* * Needed for day of the week buttons

home
edit
ring


Izaic:
@daniel-thompson Bump

Is it possible to get this merged? Looks like a insanely useful app.

daniel-thompson:
Sorry this is taking so long. I wanted to check how much extra RAM is used by the new features. We are starting to get a bit tight in that area...

Izaic:
> Sorry this is taking so long. I wanted to check how much extra RAM is used by the new features. We are starting to get a bit tight in that area...

Could this be merged but disabled by default?

daniel-thompson:
Cost looks to be 200 or so bytes for a default install (not sure from where) and goes up to 600 or so when the alarm app is enabled (and more still when it starts).

Overall this some of the best optimized code I've seen submitted for wasp-os (including most of my own early code).

Applied!

PR #234, opened 2021-07-29
apps: Fixed weather app with GadgetBridge.

taitberlette:
When I created the weather app I didn't have GadgetBridge installed, so I tried to follow the protocol on the [espurino website](https://www.espruino.com/Gadgetbridge), but it wasn't very helpful and I made some mistakes. This commit should fix these mistakes to stop the weather app from crashing, and so it displays the correct values. I have also added a new settings option called "Units", where apps can see what units the user would prefer (metric/imperial).

### DIFF
--- a/wasp/apps/settings.py
+++ b/wasp/apps/settings.py
@@ -37,7 +37,9 @@ class SettingsApp():
         self._dd = wasp.widgets.Spinner(20, 60, 1, 31, 1)
         self._mm = wasp.widgets.Spinner(90, 60, 1, 12, 1)
         self._yy = wasp.widgets.Spinner(160, 60, 20, 60, 2)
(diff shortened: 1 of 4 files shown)


taitberlette:
I added a "No weather data." message if there is no weather data available. 

Earlier, it showed the temperature as -273, and the other fields were either blank or showed a zero, which didn't look that nice.

0100001001000010:
Sorry for the late reply, as I had very bad internet for a while and then was very busy for a few days after I got good internet again. I tested out your changes, and they work well. Gadgetbridge works, as does your Android app and the simulator.

taitberlette:
Alright, thanks! This pull request is now ready to be merged.

PR #235, opened 2021-07-31
Add toolchain warning

Izaic:
Add warning about possibly building a bricked bootloader if you use your distro's toolchain instead of the recommended GNU-RM arm one.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -42,6 +42,10 @@ with pip instead:
 
     pip3 install --user -r wasp/requirements.txt
 
+.. warning::
+
+   It is not recommended to use your distro toolchain as it may result in a bricked bootloader that cannot be recovered from on a sealed device.
+
 You will also need a toolchain for the Arm Cortex-M4. wasp-os is developed and
 tested using the `GNU-RM toolchain
 <https://developer.arm.com/tools-and-software/open-source-software/developer-tools/gnu-toolchain/gnu-rm>`_


daniel-thompson:
Is the related to the recent problemsn mentioned on Arch Linux?

If so can you share (approximate if needed) the pacman command that installs most of the prerequisites? I'm not an Arch user but I'd like to try and reproduce this using podman/docker.

Izaic:
arm-none-eabi-binutils arm-none-eabi-gcc arm-none-eabi-newlib are the problem

Izaic:
Thought about it more and there's already the warning about being careful and using the pre-made bootloader. If I have some time later I'll just make some instructions on installing the proper toolchain.

daniel-thompson:
> Thought about it more and there's already the warning about being careful and using the pre-made bootloader. If I have some time later I'll just make some instructions on installing the proper toolchain.

That's somewhat ironic because I've been trying to fix it and am getting closer and closer to blaming the compiler!

I've chased it down the compilation of a single function in main.c:

```
--- main.gcc10.s	2021-08-01 10:05:58.128843911 +0100
+++ main.gcc11.s	2021-08-01 10:05:58.172843962 +0100
@@ -266,17 +266,12 @@
 softdev_mbr_init:
 	@ args = 0, pretend = 0, frame = 16
 	@ frame_needed = 0, uses_anonymous_args = 0
-	push	{r0, r1, r2, r3, r4, lr}
-	movs	r3, #0
-	strd	r3, r3, [sp, #4]
-	str	r3, [sp, #12]
+	@ link register save eliminated.
+	sub	sp, sp, #16
 	mov	r0, sp
-	movs	r3, #2
-	str	r3, [sp]
-	bl	sd_mbr_command
-	add	sp, sp, #20
+	add	sp, sp, #16
 	@ sp needed
-	ldr	pc, [sp], #4
+	b	sd_mbr_command
 	.size	softdev_mbr_init, .-softdev_mbr_init
 	.section	.text.startup.main,"ax",%progbits
 	.align	1
```

The code in question allocates a structure on the stack, initializes it and passes it into another function called sd_mbr_command(). The code generated by (the ArchLinux binaries for) gcc-11 doesn't make much sense. It allocates some stack, does not initialize it and then deallocates the stack before making a tail recursive branch to sd_mbr_command() sharing it a pointer to the uninitialized stack!

That sounds deeply broken but I do still need to strip this back to a test case to be sure I haven't missed something. However right now it is looking very much like there could bug is in the gcc-11 rather than in the bootloader. This is very unexpected... I was fulling expecting to find a bug in the bootloader when I went looking and there is still scope for me to have made a mistake which is why I want to reduce this to a test case.

PR #247, opened 2021-09-09
Add morse application

fgaz:
~~This is still missing a proper icon and screenshot~~

Signed-off-by: Francesco Gazzetta <fgaz@fgaz.me>


wasp/apps/morse.py
```diff
@@ -0,0 +1,147 @@
...
+There is a preview of the next letter at the bottom of the screen.
+
+
+.. figure:: res/MorseApp.png
+    :width: 179
+
+.. note::
+
+    Due to it's niche purpose this app is not included by default in the flash
+    images. It must be separately installed and enabled.
```
fgaz, line 21: Oh oops, this is a leftover. As discussed in IRC, I'm including it in the default image. I'll delete the note when I add the images

daniel-thompson:
You know a screenshot is just the 's' key in the simulator? Filename will be correct and in the right place so from there it is just a git add and you're done.
  

  

  

  
  
  
  
  
>

fgaz:
Okay, I did that and some minor cleanup! This should be ready for review now

fgaz:
I just noticed I was on an older micropython checkout (https://github.com/daniel-thompson/micropython/tree/048bb2da83d0cbe73a829e461bd4f22b86c282bc), updated the submodule and now the .text section of the firmware is 1kB too big :-/

I think I can shave off ~500 bytes by storing the lookup table in a different way, but I'd still be 500 bytes over the limit. Any tips? Am I doing anything particularly expensive? Or are there other apps that need optimization?

daniel-thompson:
Also looks like the autotests are failing  (not enough FLASH space).

fgaz:
I resolved the conversations and heavily optimized the letter lookup, but it's still 224 bytes over the limit. How should we proceed?

daniel-thompson:
> I resolved the conversations and heavily optimized the letter lookup, but it's still 224 bytes over the limit. How should we proceed?

For now, app will need to be removed from manifest and software app list.

Does the app load using the new apps/ directory feature?

fgaz:
It does. I'll move it there then

fgaz:
I moved it, but now it doesn't get tested anymore. Do I have to expand the tests to `apps/` too?

fgaz:
@daniel-thompson I expanded the tests

wasp/apps/morse.py
```diff
@@ -0,0 +1,148 @@
+# SPDX-License-Identifier: LGPL-3.0-or-later
+# Copyright (C) 2021 Francesco Gazzetta
+
+"""Morse translator and notepad
+~~~~~~~~~~~~~~~~~~~~~~~~~
```
daniel-thompson: This line is triggering CI failures (the underlining should line up with the d in pad)

fgaz: Fixed!

fgaz:
Great! This unblocks my other prs. I'll rebase them soon.

PR #255, opened 2021-09-20
HeartApp: draw blank line ahead of cursor

kLabz:
Helps seeing where the cursor is when displaying data over previous screen

wasp-heart-01


kLabz:
Aw, forgot about DCO while posting from web interface.. feel free to just copy/paste the 2 lines if you like the change

daniel-thompson:
Has this been tested purely on the simulator or has it also been run on a real device?

kLabz:
I have had it on my real device too since I opened this PR

kLabz:
https://user-images.githubusercontent.com/6101998/135464914-b13aa3e9-c42d-4663-91c0-106866cf4680.mp4

daniel-thompson:
Interesting. The drawing routines must have been optimized more than I thought. The reason the current graph was original implemented as it was is because drawing it 1 pixel wide was the only way to draw the lines fast enough to keep up with the sensor (nRF52832 SPI periphal kinda sucks... and then micropython-for-nrf has a few quirks that make thinks worse).

PR #256, opened 2021-09-20
Flashlight: add on/off (start off)

kLabz:
Start flashlight app as "off", with a small icon, and toggle between this and actual flashlight on tap/button.
I use flashlight app in quick ring and this prevent accidental flash. 

image
image

"off" screen with my theme:
image

### DIFF
--- a/wasp/apps/flashlight.py
+++ b/wasp/apps/flashlight.py
@@ -31,6 +31,7 @@ class TorchApp(object):
 
     def background(self):
         """De-activate the application (without losing state)."""
+        self.__activated = False
         wasp.system.brightness = self._brightness
 
     def tick(self, ticks):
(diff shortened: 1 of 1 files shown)


daniel-thompson:
I'm not sure we can allow the default to be off. That would mean extra clicking to turn the flashlight on when opening it from the  launcher.

fgaz:
Maybe there could be a way to tell if an app is launched from the launcher or from the quick ring

daniel-thompson:
> Maybe there could be a way to tell if an app is launched from the launcher or from the quick ring

That would work for me. I think `self in wasp.system.quick_ring` in the foreground() method would be sufficient to set an appropriate default.

kLabz:
Yeah i'm doing something similar in my fork, but based on big changes that wouldn't be done here. This line should be enough to reproduce a similar behavior (except that if you add it to quick ring _and_ launch it from app launcher it'll start off, but I suppose it's fine?)

PR #257, opened 2021-09-22
Add gallery app

fgaz:
An app that shows images stored in the filesystem. For example I'm using it to show a QR code containing my vCard.

gallery_icon (icon done in like 5 minutes in gimp :sweat_smile:)

GalleryApp

To do:

* [x] once it's confirmed that i did it right in #247, move the app to apps/, since there isn't enough space
* [ ] maybe revisit the icon...

Ideas:

The current format (raw rgb565), while efficient (balancing code size and image size), is not user friendly. I could add a [PNM](https://en.wikipedia.org/wiki/Netpbm#File_formats) decoder (type P6 especially). PNM is supported in programs such as GIMP so it'd be easier for an user to encode their images. On the other hand, the user already need to use the command line to invoke wasptool, so it isn't that much of an advantage.

In addition/alternative, there should be a way to decode rgb565-encoded images of variable sizes. there are a few options for that:
  * store the size in the filename
    * not great to put metadata there
    * easy to set by the user
  * store the size in a 2-byte header (plus another byte to distinguish it from PNM)
    * more resilient
    * more difficult to set by the user without a script
  * implement [PAM](https://en.wikipedia.org/wiki/Netpbm#PAM_graphics_format), with a custom RGB565 TUPLTYPE, or similar
    * theoretically the best option, but...
    * more complex to parse
    * it'd still require a custom tool


fgaz:
I moved it to apps/.

@daniel-thompson thoughts on encodings?

daniel-thompson:
> @daniel-thompson thoughts on encodings?

Not really.

It would be nice to be able to store images directly from ImageMagick or an GUI editor but unless I missed something the standard formats rule out binary RGB565. For uncompressed full screen images then using RGB888 will drops the total number of images we can fit on the filesystem from ~36 to to ~24 so I think we reall do want to store as binary RGB565.

That means a custom header and a conversion tool is probably the best route (?rename rle_encode.py to image_encode.py and add additional formats?). Maybe use a 8-byte header though, so that such files are easily detected, perhaps ('w', 'a', 's', 'p', 0, x, 0, y)?

leofidus:
For what it's worth, BMP has a [RGB565 subtype](https://en.wikipedia.org/wiki/BMP_file_format#RGB_video_subtypes), and at least imagemagick and Gimp support it.

Gimp support:
image

fgaz:
Great find, @leofidus!

BMP is pretty simple ...except for the million different headers, but I guess we can just implement V5, which seems to be what GIMP and ImageMagick use (though GIMP uses V3 if you check the "Do not write color space information" checkbox).

fgaz:
BMP works in the simulator, but as espected it goes out of memory on the watch... 1.6K of bytecode is too much. Maybe I could remove the icon or the header checks, but that wouldn't be very user friendly

fgaz:
When frozen in micropython.zip, the app works well! I even added some fancier error reporting.

The only downside is that bmp is slower to render, probably due to the byte swapping that has to be done because of the endianness

PR #272, opened 2021-11-03
Make all notifications show in pager instead of one at a time 

nabakolu:
The way notifications are currently shown makes it really hard to get an overview, because only one notification is shown at a time and it gets removed after seeing it.

This patch makes a single string of all notifications and shows them in the pager.
Viewing the notifications does not remove them, but the ability to remove all notification is still available.  

NotificationsApp

### DIFF
--- a/wasp/apps/pager.py
+++ b/wasp/apps/pager.py
@@ -88,10 +88,15 @@ class NotificationApp(PagerApp):
 
     def foreground(self):
         notes = wasp.system.notifications
-        note = notes.pop(next(iter(notes)))
-        title = note['title'] if 'title' in note else 'Untitled'
-        body = note['body'] if 'body' in note else ''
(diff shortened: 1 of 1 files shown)


daniel-thompson:
Hi @nabakolu 

I'm pleased you been able to mod this to be the way you like it. However I'm afraid I don't like this UI concept.

For notifications wasp-os views the watch as an auxilliary to the phone... when the watch buzzes I expect people to read (and clear) the notification soon after receiving it. That's why reading it on the watch will not clear it from the phone.

So the model is (or will be) that you read and dismiss on the watch because the master copy will be on your phone as a reminder if needed. Currently we have the (broken) concept of clear all notifications but this is only a workaround for the fact that clearing notifications on the phone doesn't yet clear them from the watch. Once that is fixed then if you have 6 pending notifications on the watch it would be better to get out your phone and clear them there.

PR #274, opened 2021-11-19
Add *.mpy to gitignore

fgaz:
.mpy files are micropython bytecode and should be ignored.
Now that user-defined apps are implemented, it's likely that .mpy files
will be present in apps/


daniel-thompson:
Merged. Thanks!

PR #275, opened 2021-11-20
Add level app

fgaz:
A simple level that can be calibrated

level_icon
LevelApp
LevelApp

TODO:

* [x] documentation

Closes #187


apps/Level.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def tick(self, ticks):
+        self._update()
+        wasp.system.keep_awake()
+
+    def touch(self, event):
+        if self.prompt:
+            # Handle buttons
+            if self.calibrate.touch(event):
+                (y, x, _) = watch.accel.accel_xyz()
```
daniel-thompson, line 102: Like #277 this means that the orientation of the accelerometer ends up burned into the apps.

Ideally the driver layer would handle the remap so that the x/y on the accelerometer would match x/y on the display (in other words the x value should be positive if the left side of the watch frame is higher than the right).

It's actually a little awkward to solve though since the BMA421 driver can't know what orientation the unit is within the watch frame either so the BMA421 driver would need to be instanciated with a remapping value (I suspect a single 32-bit value is more than enough to describe the mapping and sign between sensor x/y/z and frame x/y/z).

README.rst
```diff
@@ -257,3 +257,7 @@ application (and the "blank" white screen is a torch application):
 .. image:: res/WordClkApp.png
    :alt: Shows a time as words in the wasp-os simulator
    :width: 179
+
+.. image:: res/LevelApp.png
+   :alt: Shows a time as words in the wasp-os simulator
```
thiswillbeyourgithub, line 262: @fgaz I think you forgot to update this string :)

PR #277, opened 2021-11-20
Add beacon app

fgaz:
Flash the relatively powerful HRS led repeatedly, mostly for signaling purposes.

The sliders are for frequency and intensity.

The blinking is handled by the HRS, so this app consumes very little power.
With BLE and/or step counter disabled and blinking frequency set to the minimum,
the watch's battery will last for many days.

Applications: marking a spot in the night, emergency bike rearlight... having fun with the led ;)

beacon_icon

BeaconApp

TODO:

* [x] documentation
* [x] improve the interface
  * [x] status bar
  * [x] slider labels


fgaz:
This is now ready for review

apps/Beacon.py
```diff
@@ -0,0 +1,130 @@
...
+        if self._slider_current.value % 2:
+            self._reg_pdriver |= _PDRIVER_PDRIVE0
+        else:
+            self._reg_pdriver &= ~_PDRIVER_PDRIVE0
+        if self._slider_current.value >> 1:
+            self._reg_enable |= _ENABLE_PDRIVE1
+        else:
+            self._reg_enable &= ~_ENABLE_PDRIVE1
+        self._reg_enable = (self._reg_enable & (~_ENABLE_HWT)) | (self._slider_wait_time.value << 4)
+        wasp.watch.hrs.write_reg(_PDRIVER, self._reg_pdriver)
```
daniel-thompson, line 122: I don't think direct calls from apps to read/write registers are the best way to do this. In general we want to reduce the number of places hrs3300 specific assumptions get burnt into apps.

Might be better to have an method to specify the LED intensity.

fgaz: Addressed

PR #280, opened 2021-11-30
Add DisaBLE app for disabling BLE

fgaz:
Small application for disabling bluetooth to save power and enhance security

To re-enable it, it uses the reset hack as discussed on IRC

I had to draw a small monochrome icon and optimize calc to make space for this app.

disaBLE_icon
DisaBLEApp


fgaz:
weird... I can build the image with `make pinetime` but CI fails with "region `FLASH_TEXT' overflowed by 464 bytes". Maybe it's the gcc version? I'm using 10.3.0 while it looks like CI uses version 9-2019-q4

daniel-thompson:
> weird... I can build the image with `make pinetime` but CI fails with "region `FLASH_TEXT' overflowed by 464 bytes". Maybe it's the gcc version? I'm using 10.3.0 while it looks like CI uses version 9-2019-q4

I have switched over to the 10-2020-q4 toolchain (I know its already out-of-date but its the one I hve been testing on for a long time and is supported by the CI actions).

PR #285, opened 2021-12-31
Add weekday to default watchface

fgaz:
It fits nicely and I find it very useful


daniel-thompson:
This has come up before (in #147) but I like the clock without the day of the week (just as I like it to be 24 hours).

These things are better implemented additively. We have all the machinary to select different watch faces so we can register multiple digital clocks.

Given we are short of flash space there is some other tidying up we could do to make enough space to do this.

1. Remove all icons from all watchface apps (icon.clock plus the custom ones in some of the clock apps).
2. Modify clock.py slightly so that date string is drawn from a sub-method
3. Inherit a new digital clock from the existing one and override the method to draw the date.

PS Sorry I didn't mention #1 above in IRC the other day... it has only just occured to me!

fgaz:
> Remove all icons from all watchface apps (icon.clock plus the custom ones in some of the clock apps).

icon.clock is also used in the watchface chooser. should I remove it anyway?

fgaz:
Done!

PR #297, opened 2022-01-24
Correct HRM gain value

kieranc:
The current value of 0x03 sets the gain to 1x, this change should correctly set it to 8x.

Signed-off-by: Kieran Cawthray <kieranc@gmail.com>

### DIFF
--- a/wasp/drivers/hrs3300.py
+++ b/wasp/drivers/hrs3300.py
@@ -43,7 +43,8 @@ class HRS3300:
 
         # 64x gain
         #w(_HGAIN, 0x10)
-        w(_HGAIN, 0x03)
+        # 8x gain
+        w(_HGAIN, 0xc)
 
     def read_reg(self, addr):
         return self._i2c.readfrom_mem(_I2CADDR, addr, 1)[0]


fgaz:
`0x03` actually maps to 8x (see the datasheet at https://wiki.pine64.org/wiki/PineTime#Component_Datasheets )

kieranc:
As I understand, 0x3 becomes 00000011 and only bits 2-4 are used to set the gain, so 0x3 is 000 for bits 2-4 which is 1x gain.

image

fgaz:
Huh, you're right the alignment is wrong!

You can fix the DCO by making the author field of the commit and the signoff line match:

```
git commit --amend --author='Kieran Cawthray <kieranc@gmail.com>'
```

Also I'd write `0xc` as `0x0c` to make it clear that there are 8 bits

daniel-thompson:
> Also I'd write `0xc` as `0x0c` to make it clear that there are 8 bits

Perhaps the best way to right this is perhaps 3 << 2 (since that would make it less likely to same problem reoccurs if we ever decide to swap to x8).

Also, has anyone tried this change for real (IIUC it was discovered by inspection... and many thanks for the PR if that is the case). A test report from someone running on the real watch would be great (naturally I will if nobody else does... but it means we may have to wait a bit longer before we merge it).

kieranc:
> > Also I'd write `0xc` as `0x0c` to make it clear that there are 8 bits
> 
> Perhaps the best way to write this is perhaps 3 << 2 (since that would make it less likely to same problem reoccurs if we ever decide to swap to x4).
> 
> Also, has anyone tried this change for real (IIUC it was discovered by inspection... and many thanks for the PR if that is the case). A test report from someone running on the real watch would be great (naturally I will if nobody else does... but it means we may have to wait a bit longer before we merge it).

In a wonderfully circular approach, Infinitime was going to use 8x since it had been successful for wasp-os... Some experimentation and user reports also looks like 8x provides the best data but it hasn't been extensively tested.

image

As far as this PR/DCO is concerned, I opened the PR to bring the issue to your attention but I'm not worried about credit if someone more familiar with your codebase wants to make the change instead. I'm slightly ashamed to admit I didn't even fork the repo, I just made the edit via the github website so amending the commit would be a bit of a pain.

fgaz:
> Perhaps the best way to write this is perhaps 3 << 2 (since that would make it less likely to same problem reoccurs if we ever decide to swap to x4).

maybe even `0b011 << 2`

> A test report from someone running on the real watch would be great

I just cherry-picked this in my fork and will try it soon

fgaz:
Well... I guess it isn't worse? But HRMs never worked well on me so I trust the graph above more and this is a +1 from me.

Issue #303, opened 2022-02-23
error when wasptool --pull cannot create local directory or file


thiswillbeyourgithub:
Hi,

As mentioned earlier today in the matrix room, when using `wasptool --pull A/B/C`, if creating local directory `A` or `A/B` fails, the error message will not be comprehensible.

I think it would be worth it to add a tiny test to write to the local file before pulling. If this test fails, print a more explicit Exception.

Have a nice day!

PR #304, opened 2022-02-24
fix: wasptool --pull fails when downloading nested file

thiswillbeyourgithub:
fix to : #303

### DIFF
--- a/tools/wasptool
+++ b/tools/wasptool
@@ -245,6 +245,9 @@ def check_rtc(c):
 def handle_binary_download(c, tname, fname):
     verbose = bool(c.logfile)
 
+    if tname.startswith("/"):
+        fname = "." + fname
+
     c.run_command('import os')
     stat = c.run_command(f'os.stat("{tname}")[6]')
     if 'Error' in stat:
--- a/tools/wasptool
+++ b/tools/wasptool
@@ -6,7 +6,7 @@
 import argparse
 import io
 import random
-import os.path
+import os
 import pexpect
 import time
 import types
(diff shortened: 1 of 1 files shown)


tools/wasptool
```diff
@@ -258,6 +258,25 @@ def handle_binary_download(c, tname, fname):
...
+                if local_dir == ".":
+                    continue
+                try:
+                    os.mkdir(f"{pile}/{local_dir}/")
+                    print(f"Created local dir {pile}/{local_dir}/")
+                except FileExistsError:  # folder already exists
+                    pass
+                pile += local_dir + "/"
+        with open(fname, 'wb') as f:
+            f.write(b"write_test")
```
fgaz, line 274: Writing to the file is probably a bad idea. What if it contains important data? I'd just text for existence instead

tools/wasptool
```diff
@@ -258,6 +258,25 @@ def handle_binary_download(c, tname, fname):
...
+    except:
+        new_name = str(int(time.time())) + ".file"
+        print(f"Error when test writing to {fname}, downloading instead to local file '{new_name}'")
+        fname = new_name
```
fgaz, line 278: Technically there could already be a file named like that.

Unique filenames are usually generated using https://docs.python.org/3/library/tempfile.html

tools/wasptool
thiswillbeyourgithub: I felt the same way but followed logic of the original file : if you already had a fname file it would be overwritten.

If you still think it's better just tell me and I'll do it.

tools/wasptool
thiswillbeyourgithub: I am not familiar with tempfile so I'm not comfortable trying this.

While you are correct, this ugly workaroud seems to me vastly better than the current state. Otherwise I could just raise an exception. Your call.

tools/wasptool
thiswillbeyourgithub: I took your input into consideration and redid a new commit.

thiswillbeyourgithub: I took your input into consideration and redid a new commit.

tools/wasptool
```diff
@@ -364,6 +382,8 @@ if __name__ == '__main__':
             help='Report on the current memory usage.')
     parser.add_argument('--pull',
             help='Fetch a file from the target')
+    parser.add_argument('--force', action='store_true', dest='force_pull',
```
thiswillbeyourgithub, line 385: I do think that the order of the arguments (alphabetical) is suboptimal and should instead be grouped by topic.

"Force" should be close to "pull" and "as" should be close to "push".

thiswillbeyourgithub:
image
So far it seems to work flawlessly

PR #306, opened 2022-02-24
feat: add repeat / chime feature to timer app

thiswillbeyourgithub:
TODO :
* ~~if this gets approved, this page should be updated : https://wasp-os.readthedocs.io/en/latest/apps.html#timer-application~~
* ~~without forgetting to update the screenshots~~


wasp/apps/timer.py
```diff
@@ -14,7 +14,6 @@
 
 import wasp
 import fonts
-import time
```
thiswillbeyourgithub, line 17: this import was not needed so I removed it

thiswillbeyourgithub:
The reason the test is failing is `overflowed by 124 bytes`. This could be maybe solved by the PR that merges disaBLE to the settings?

daniel-thompson (changes requested):
A patch description explaining what this change does and why would be really useful.

In general what is the purpose of the "Repeat" checkbox? In other words why does the user need to make a decision. Can't we just pick the "best" behaviour all the time?

daniel-thompson:
> The reason the test is failing is `overflowed by 124 bytes`. This could be maybe solved by the PR that merges disaBLE to the settings?

Let's see what happens when this comes back for re-testing. It's possible replacing Game of Life with the Gallery app will have created some extra space.

fgaz:
While the chime feature is great, I often like my watch to be as silent as possible. Maybe the chime could be optional?

Or even better, there could be a global setting that has 3 volume options (completely silent, vibration, sound) that applies to _all_ apps (well, the timer and the alarm)

thiswillbeyourgithub:
> While the chime feature is great, I often like my watch to be as silent as possible. Maybe the chime could be optional?
> 
> Or even better, there could be a global setting that has 3 volume options (completely silent, vibration, sound) that applies to _all_ apps (well, the timer and the alarm)

I don't understand. To me, the goal of the timer app is to be notified that X time has passed. I don't see the point of a silent timer app.

thiswillbeyourgithub:
I'm thinking of working on a pomodoro timer app, largely based on this fork. Maybe you would prefer having two separate apps and leaving this one as is?

thiswillbeyourgithub:
In the end I don't think this PR is a good idea, seeing the progress i made in #320

PR #309, opened 2022-02-25
add code timer

thiswillbeyourgithub:
- add code_timer.py
- docs: mention code_timer.py

Thought this could be useful to other devs

### DIFF
--- a/tools/code_timer.py
+++ b/tools/code_timer.py
@@ -1,9 +1,11 @@
 # SPDX-License-Identifier: LGPL-3.0-or-later
 # Copyright (C) 2022 thiswillbeyourgithub user of github.com
-# code_timer.py
-# put the code you want to compare in the content of f1 and f2
-# then use './wasptool --exec code_timer.py'
-# to see which is faster on the watch
+
+"""
+put the code you want to compare in the content of f1 and f2
+then use './wasptool --exec code_timer.py'
+to see which is faster on the watch
+"""
 
 import wasp
 wasp.gc.collect()
--- a/tools/code_timer.py
+++ b/tools/code_timer.py
@@ -1,0 +1,34 @@
+# code_timer.py
+# put the code you want to compare in the content of f1 and f2
+# then use './wasptool --exec code_timer.py'
+# to see which is faster on the watch
+
+import wasp
+wasp.gc.collect()
+t = wasp.machine.Timer(id=1, period=8000000)
+
+# example : comparing pow(5, 5) with 5**5
+from math import pow
+
+def f1():
+    return 5**5
+def f2():
+    return pow(5, 5)
+
+vals1 = []
+vals2 = []
+a = 0
+for i in range(100):
+    t.start()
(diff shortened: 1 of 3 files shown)


tools/code_timer.py
```diff
@@ -0,0 +1,34 @@
+# code_timer.py
```
daniel-thompson, line 1: Needs a SPDX-license-identifier comment at the top.

daniel-thompson, line 1: I wasn't quite sure what a "code timer" is. Perhaps just call it benchmark?

tools/code_timer.py
```diff
@@ -0,0 +1,34 @@
...
+a = 0
+for i in range(100):
+    t.start()
+    a = f1()
+    vals1.append(t.time())
+    t.stop()
+    t.start()
+    a = f2()
+    vals2.append(t.time())
+    t.stop()
```
daniel-thompson, line 29: I wonder about having a function to run the benchmark and using accumulateld time rather than capturing everything in an array. Something like:

~~~
def benchmark(fn, iterations=100):
    t.start()
    for i in range(iterations):
         pass
    overhead = t.time()
    t.stop()
    t.start()
    for i in range(iterations):
        fn()
    elapsed = t.time()
    t.stop()
    return (elapsed - overhead) / iterations
~~~

tools/code_timer.py
thiswillbeyourgithub: fixed in [413b4e3](https://github.com/daniel-thompson/wasp-os/pull/309/commits/413b4e36aab2c27babcd8896d4e4946a78e3cc75)

thiswillbeyourgithub: fixed in f05521cc0cc08cf2d03c2459535dc22ae8f0f5db

tools/code_timer.py
thiswillbeyourgithub: Well the overhead is the same for both function so I don't think it's relevant to substract it.

thiswillbeyourgithub:
Sorry, for the number of commits, I ended up changing quite a bit of code in the last commit because I kept having memory errors despite resetting the watch.

tools/code_timer.py
daniel-thompson: Why not? The loop overhead is not part of what is being benchmarked and by removing the overhead we can make it more complex without hurting the benchmark results (for example we could add min/max tracking).

There are a couple of other reasons to keep the timers running during the loop.

1. Vastly lower memory overhead (even the newer array based implementation still needs 800 bytes to hold the information).
2. Local variables are deleted automatically so we don't need any del commands.
3. Having it as function makes the usage more pythonic then hand editing a script and relaunching it.

BTW by pythonic I mean the usage (after copying the library code to SPINOR so we can import it) looks like normal python interaction at the REPL:

>>> import code_benchmark
>>> benchmark(one_approach)
100
>>> benchmark(another_approach)
105
>>>

PR #310, opened 2022-02-28
integrate disaBLE app in settings app

thiswillbeyourgithub:
Signed-off-by: thiswillbeyourgithub <github@32mail.33mail.comm>

I hope it's fine with you @fgaz. I was just thinking it might be better to put disaBLE in the settings for now.

What do you think?


fgaz:
iirc we decided to go with a separate app because it's a bit unfinished (can't turn ble off without rebooting). @daniel-thompson?

wasp/boards/manifest_240x240.py
```diff
@@ -7,7 +7,6 @@
     'apps/calc.py',
     'apps/clock.py',
     'apps/chrono.py',
-    'apps/disaBLE.py',
```
fgaz, line 10: there are a bunch of other places where you need to remove references to DisaBLE:

```
$ git grep '[dD]isaBLE'
README.rst:.. image:: res/DisaBLEApp.png
docs/apps.rst:.. automodule:: apps.disaBLE
wasp/apps/disaBLE.py:"""DisaBLE
wasp/apps/disaBLE.py:.. figure:: res/DisaBLEApp.png
wasp/apps/disaBLE.py:class DisaBLEApp():
wasp/apps/disaBLE.py:    NAME = 'DisaBLE'
wasp/apps/disaBLE.py:    # 1-bit RLE, 96x64, generated from res/disaBLE_icon.png, 167 bytes
wasp/apps/software.py:        db.append(('disaBLE', factory('DisaBLE')))
wasp/boards/manifest_240x240.py:    'apps/disaBLE.py',
```

Also the `res/DisaBLEApp.png` and `res/disaBLE_icon.png` files

thiswillbeyourgithub: Thanks!

daniel-thompson:
Agree with @fgaz . This is deliberately implemented as an app because the feature is currently a bit hacky.

fgaz:
By the way I think the infinitime people managed to turn bluetooth back on without rebooting... we should investigate how they do it

JF002:
InfiniTime person here :)
The airplane mode in InfiniTime does not completely turn the bluetooth radio OFF. Instead, it disconnects from the device that is currently connected and it disables advertisements. This way, the watch should be mostly invisible and should not emit any signal in the air. I don't think anyone has actually checked that it's completely silent, though.

fgaz:
Also I just remembered infinitime uses a different bluetooth stack, so we can't really transfer knowledge between the projects :-/

PR #313, opened 2022-02-28
clarify doc on sleep

thiswillbeyourgithub:
Note that I am not familiar with readthedocs nor graphviz so was not able to recompile the documentation.

### DIFF
--- a/docs/appguide.rst
+++ b/docs/appguide.rst
@@ -70,8 +70,8 @@ any change in its lifecycle.
         BACKGROUND -> START [ label=" __del__()   " ];
         BACKGROUND -> ACTIVE [ label=" foreground()   " ];
         ACTIVE -> BACKGROUND [ label=" background()   " ];
-        ACTIVE -> GO_TO_CLOCK [ label=" sleep() -> False   " ];
-        GO_TO_CLOCK -> BACKGROUND [ label=" background()   " ];
+        ACTIVE -> RET_TO_CLOCK [ label=" sleep() -> False   " ];
+        RET_TO_CLOCK -> BACKGROUND [ label=" background()   " ];
         ACTIVE -> SLEEPING [ label=" sleep() -> True   " ];
         SLEEPING -> ACTIVE [ label=" wake()   " ];
 
@@ -79,7 +79,7 @@ any change in its lifecycle.
         BACKGROUND [ shape=box, style=rounded ]
         ACTIVE [ shape=box, style=rounded ]
         SLEEPING [ shape=box, style=rounded ]
-        GO_TO_CLOCK [ label="GOTO ClockApp" ];
+        RET_TO_CLOCK [ label="Return to ClockApp" ];
     }
 
 When an application is initialized is enters the ``BACKGROUND`` state. A
--- a/docs/appguide.rst
+++ b/docs/appguide.rst
@@ -97,31 +97,38 @@ A backgrounded application enters the ``ACTIVE`` state via a call to
 :py:meth:`~.TemplateApp.foreground`. When it is active the application owns the
 screen and must draw and maintain its user interface.
 
-If the system manager wants to put the watch to sleep then it will tell the
-active application to :py:meth:`~.TemplateApp.sleep`.
-If the application returns True then the application will remain active
-whilst the watch is asleep.
-It will receive no events nor the application tick whilst the system is
-asleep and, instead, must wait for a :py:meth:`~.TemplateApp.wake` notification
-telling the application that the device is waking up and that it may
-update the screen if needed.
+If the screen timeout is near, the system manager will want to turn the screen
+off to save power. It will call the sleep entry point of the currently 
+active application (here, it would be :py:meth:`~.TemplateApp.sleep`).
+Several situations are possible:
 
-If an application does not support sleeping then it can simply not implement
-:py:meth:`~.TemplateApp.sleep` or :py:meth:`~.TemplateApp.wake`.
-In this case the system manager will automatically return to the default
-application, typically the main clock face.
+* If no sleep() method was implemented in the app, a default sleep()
+  method will be used to turn the screen off, put the application in the
+  background and switch to the default application (usually the main clock
+  application). Turning the screen on will appear as though the user
+  exited the application.
+* If the sleep() method was implemented and returns True, turning the
+  screen back on will directly show the app instead of the home menu.
+* If the sleep() method returns False, turning the screen on will show
+  the main clock application.
 
-Some applications may support sleeping only under certain circumstances. For
-example a stopwatch may choose to remain active when the watch sleeps only if
-the stopwatch is running.
-This type of application must implement :py:meth:`~.TemplateApp.sleep` and
-return False when it does not want to remain active when the system
-resumes.
+Here, "sleep" only refers to the appearance of the watch (screen turned off,
+unresponsive to touch) and not to any kind of power saving feature. For
+example, sleeping has no effect on bluetooth connectivity.
+
+Note that while the system is asleep (no matter what method were
+implemented, if any), the application cannot receive events
+nor ticks. Instead, it must wait for a :py:meth:`~.TemplateApp.wake`
+notification telling the application that the device is waking up,
+allowing it to update the screen if needed.
 
 .. note::
 
     Most applications should not implement :py:meth:`~.TemplateApp.sleep`
     since it is often a better user experience for the watch to return to the
+    default application when they complete an interaction. Notable
+    exceptions include clock related apps like the stopwatch, which returns
+    True when the stopwatch is running and False instead.
 
 The full list of application entry points can be found at the bottom of this page.
 
@@ -151,7 +158,7 @@ event notifications and timer callbacks:
   and specify the tick frequency.
 
 Additionally if your application is a game or a similar program that should
-not allow the watch to go to sleep when it is running then it should
+not allow the watch to go to sleep (turning off the screen) when it is running then it should
 arrange to call :py:meth:`~.Manager.keep_awake` from the application's
 :py:meth:`~.TemplateApp.tick` method.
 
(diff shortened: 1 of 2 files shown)


docs/appguide.rst
```diff
@@ -68,16 +70,16 @@ any change in its lifecycle.
...
+        RET_TO_CLOCK -> BACKGROUND [ label=" background()   " ];
+        ACTIVE -> SYSTEM_SLEEPING [ label=" sleep() -> True   " ];
+        SYSTEM_SLEEPING -> ACTIVE [ label=" wake()   " ];
 
         START [ shape=box ];
         BACKGROUND [ shape=box, style=rounded ]
         ACTIVE [ shape=box, style=rounded ]
-        SLEEPING [ shape=box, style=rounded ]
-        GO_TO_CLOCK [ label="GOTO ClockApp" ];
+        SYSTEM_SLEEPING [ shape=box, style=rounded ]
```
daniel-thompson, line 81: This diagram is the an application state diagram not a system state diagram. In other words it diagrams the state an app can be in and what app entry points are called to change application state.

SYSTEM_SLEEPING is not a valid state for an application to be in (an app is an app, the system is the system, only the system can be in a system sleep state).

docs/appguide.rst
```diff
@@ -68,16 +70,16 @@ any change in its lifecycle.
...
+        RET_TO_CLOCK [ label="Return to ClockApp" ];
```
daniel-thompson, line 82: I'm afraid I'm also not a fan of this rename either. States are UPPERCASE in this diagram (to distinguish them from the edges which are app API entry points). Also "Return" implies some kind of backstack (like Android) that wasp-os doesn't have.

docs/appguide.rst
```diff
@@ -95,32 +97,40 @@ A backgrounded application enters the ``ACTIVE`` state via a call to
...
-application, typically the main clock face.
-
-Some applications may support sleeping only under certain circumstances. For
-example a stopwatch may choose to remain active when the watch sleeps only if
-the stopwatch is running.
-This type of application must implement :py:meth:`~.TemplateApp.sleep` and
-return False when it does not want to remain active when the system
-resumes.
+If the screen timeout is near, the system manager will want to turn the screen
+off to save power. It will call the sleep entry point of the currently
```
daniel-thompson, line 101: I don't think it is a clarification to replace "put the watch to sleep" with "turn the screen off to save power". During a sleep we currently turn off backlight, display controller, touch panel and CPU. In the future we may also want to turn off the BLE advertisements too if there is not a currently active bluetooth connection.

docs/appguide.rst
```diff
@@ -95,32 +97,40 @@ A backgrounded application enters the ``ACTIVE`` state via a call to
...
+If the screen timeout is near, the system manager will want to turn the screen
+off to save power. It will call the sleep entry point of the currently 
+active application (here, it would be :py:meth:`~.TemplateApp.sleep`).
+Several situations are possible:
+
+* If no sleep() method was implemented in the app, a default sleep()
+  method will be used to turn the screen off, put the application in the
+  background and switch to the default application (usually the main clock
+  application). Turning the screen on will appear as though the user
+  exited the application.
```
daniel-thompson, line 109: It would be better to describe this as two possible cases:

1. App implements :py:meth:`~/TemplateApp.sleep()` and returns True
2. Either the app does not implement :py:meth:`~/TemplateApp.sleep()` or the app implements :py:meth:`~/TemplateApp.sleep()` and it returns False.

Also I am not clear why we need to say what happens when we leave the low power state. It is clearer to say what does happen (the application is closed by the system during the system sleep)Does it really appear as though the user exited the application when the ...

wasp/apps/template.py
```diff
@@ -44,23 +44,30 @@ def __init__(self):
...
+        about to be turned off. Most apps should not implement it.
+        If not implemented or returning True: goes back to the default clock
+        application. If implemented and returning False: turning the screen
+        back on will show the current application"""
         return False
 
     def wake(self):
-        """Notify the application the device is waking up."""
+        """Notify the application that the screen is waking up. Generally used
+        to redraw the screen or to re-enable ticks."""
```
daniel-thompson, line 70: I thought the tick was re-enabled automatically by the system.

docs/appguide.rst
```diff
@@ -49,6 +49,8 @@ Some of the key points of interest in this example application are:
    a common wasp-os idiom that is normally used to pattern to distinguish a full
    refresh of the screen from an fast update (a redraw).
 
+The full list of application entry points can be found at the bottom of this page.
+
```
daniel-thompson, line 53: Cross-linking here is wrong.

This material is strictly introductory and the user is not yet ready to make a decision to follow this cross-link or not.

Instead, if the API entry points are the next thing we think the reader should read then that section should be physically moved here. If the Application life-cycle section is better next material to read then we don't need a cross-link.

Cross links are important to navigating a document but they should never be a "workaround" for a document that is simply in the wrong order!

daniel-thompson (commented):
I'm afraid this review has raised a lot of things I don't like. Don't worry. These things happen!

I think the short version is that I mostly like the changes to expand the (extremely short) description of the purpose of the API entry points (in templateapp.py) but I'm not a fan of most of the proposed changes to the application lifecycle. I'm not convinced they are significantly clearer, just differently phrased.

docs/appguide.rst
thiswillbeyourgithub: Alright, what's the best way to remove this? Should I just do a git revert ?

docs/appguide.rst
thiswillbeyourgithub: Alright, I'll remove it

docs/appguide.rst
thiswillbeyourgithub: Ah shoot, I was convinced that sleeping only changes wether the app will return to the home menu or not.

docs/appguide.rst
thiswillbeyourgithub: I moved the Application entry point list higher up

wasp/apps/template.py
thiswillbeyourgithub: I removed the section about tics then

docs/appguide.rst
thiswillbeyourgithub: I just added your text in 78bb996

docs/appguide.rst
```diff
@@ -95,32 +97,45 @@ A backgrounded application enters the ``ACTIVE`` state via a call to
...
-
-Some applications may support sleeping only under certain circumstances. For
-example a stopwatch may choose to remain active when the watch sleeps only if
-the stopwatch is running.
-This type of application must implement :py:meth:`~.TemplateApp.sleep` and
-return False when it does not want to remain active when the system
-resumes.
+If the screen timeout is near, the system manager will want to turn the screen
+off to save power. It will call the sleep entry point of the currently 
+active application (here, it would be :py:meth:`~.TemplateApp.sleep`).
```
daniel-thompson, line 102: The text in brackets is too verbose. Just make the usage of sleep on line 101 into a link.

docs/appguide.rst
daniel-thompson: The phrasing can be improved but it should focus on user experience.

~~~
If the user does not interact with the device for a period of time then the system manager will put the device to sleep and turn off the screen to save power. It will call the :py:meth:`~.TemplateApp.sleep` entry point of the currently active application to notify the app that the device will shortly go to sleep.

docs/appguide.rst
```diff
@@ -95,32 +97,45 @@ A backgrounded application enters the ``ACTIVE`` state via a call to
...
+
+
+* App implements :py:meth:~/TemplateApp.sleep() and returns True
+* Either the app does not implement :py:meth:~/TemplateApp.sleep()
+  or the app implements :py:meth:~/TemplateApp.sleep() and it returns False.
+
+
+Here, "sleep" only refers to the appearance of the watch (screen turned off,
+unresponsive to touch) and not to any kind of power saving feature. For
+example, sleeping has no effect on bluetooth connectivity.
```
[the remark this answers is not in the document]
daniel-thompson: Power saving is complex (AFAIR there can be different running power states for BLE). I certainly don't like "not any kind of power saving". If you want to reassure the reader that pre-existing BLE connections remain functional whilst the device is asleep then let say that.

docs/appguide.rst
```diff
@@ -148,8 +163,8 @@ event notifications and timer callbacks:
   and specify the tick frequency.
 
 Additionally if your application is a game or a similar program that should
-not allow the watch to go to sleep when it is running then it should
-arrange to call :py:meth:`~.Manager.keep_awake` from the application's
+not allow the watch to turn off the screen while running (i.e. go to sleep)
```
daniel-thompson, line 166: If you really want to clarify what "go to sleep" means then I can tolerate tasteful replacement of "go to sleep" with "go to sleep or turn of the screen".

wasp/apps/template.py
```diff
@@ -44,23 +44,30 @@ def __init__(self):
...
     def background(self):
-        """De-activate the application."""
+        """Called when the leaving the application, either because of screen
+        timeout of directly by the user."""
         pass
 
     def sleep(self):
-        """Notify the application the device is about to sleep."""
+        """Called by the system to notify the application that the screen is
+        about to be turned off. Most apps should not implement it.
```
daniel-thompson, line 62: Here I want to keep the "go to sleep" to be preserved. Although "go to sleep and turn off the screen" might be OK.

wasp/apps/template.py
```diff
@@ -44,23 +44,30 @@ def __init__(self):
...
+        If not implemented or returning True: goes back to the default clock
+        application. If implemented and returning False: turning the screen
+        back on will show the current application"""
         return False
 
     def wake(self):
-        """Notify the application the device is waking up."""
+        """Notify the application that the screen is waking up. Generally used
```
daniel-thompson, line 69: Drop "the screen" here.

docs/appguide.rst
daniel-thompson: Let's remove this paragraph. It's no longer correct (since it is not at the bottom of the page) and with the new ordering it is not useful to cross-link here.

daniel-thompson (changes requested):
Sorry... forgot to tag the "Request changes" box

PR #317, opened 2022-03-15
new: vibrate in case of exception except if in silent

thiswillbeyourgithub:
modified:   wasp/wasp.py


wasp/wasp.py
```diff
@@ -522,10 +522,14 @@ def run(self, no_except=True):
                 self._tick()
             except KeyboardInterrupt:
                 raise
-            except MemoryError:
-                self.switch(PagerApp("Your watch is low on memory.\n\nYou may want to reboot."))
+            except MemoryError as e:
+                if self.notify_level >= 2:
+                    watch.vibrator.pulse()
+                self.switch(PagerApp("Your watch is low on memory.\n\nYou may want to reboot: {}".format(str(e))))
```
fgaz, line 528: what if the .format(str(e)) calls also go oom?

thiswillbeyourgithub: What do you suggest? I feel like it's too bad that the error message get lost. I'm often having problems troubleshooting things on the pinetime.

fgaz: Honestly I'm not sure... I know that in lower level languages the memory necessary for this kind of error reporting is preallocated. Maybe here we can just add _another_ `try` around those three lines, with the `except` being a static message like the current one

PR #318, opened 2022-03-16
fix: smoothen battery level before showing it to user

thiswillbeyourgithub:
Signed-off-by: thiswillbeyourgithub <github@32mail.33mail.comm>

edit : I reworked the code to smoothen the voltage instead of the level

### DIFF
--- a/wasp/drivers/battery.py
+++ b/wasp/drivers/battery.py
@@ -26,6 +26,7 @@ class Battery(object):
         self._battery = ADC(battery)
         self._charging = charging
         self._power = power
+        self._cache = set()
 
     @micropython.native
     def charging(self):
@@ -33,7 +34,7 @@ class Battery(object):
 
         :returns: True if the battery is charging, False otherwise.
         """
-        return self._charging.value() 
+        return self._charging.value()
 
     def power(self):
         """Check whether the device has external power.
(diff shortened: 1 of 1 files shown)


wasp/drivers/battery.py
```diff
@@ -50,26 +51,34 @@ def voltage_mv(self):
...
         """
         raw = self._battery.read_u16()
-        return (2 * 3300 * raw) // 65535
+        mv = (2 * 3300 * raw) // 65535
+
+        if mv not in self._cache:
+            self._cache.add(mv)
+            while len(self._cache) > 2:
+                self._cache.remove(max(self._cache))
+        return min(self._cache)
```
fgaz, line 68: won't this always keep the minimum voltage?
for example
|loop|mv|cache|min|
|-|-|-|-|
|0|1|{1}|1|
|1|2|{1,2}|1|
|2|5|{1,2}|1|
|...|...|{1,2}|1|

thiswillbeyourgithub:
I think you're right. Indeed in some cases the battery level is stuck at the lowest possible level until I reboot but I thought it was related to another change. I'm now trying with using `.pop()` to remove the oldest value and returning the mean of the `_cache` and that should be better. Should I open a new PR for this?

fgaz:
Does the voltage reading really fluctuate so much that it needs smoothing? In my experience it's only noticeable if you look at the raw mv value.

thiswillbeyourgithub:
Personally I use the mV value and the smoothing is necessary.

thiswillbeyourgithub:
Shoot I just realized that .pop() in sets is not implementd the same way in python as in micropython. Apparently this always pops the last value instead of the oldest. I reverted to using an array in my own fork and will probably do a PR in the future

thiswillbeyourgithub:
Sorry everyone!

I noticed this some time ago so I PR'd in #387  my current version of the drivers/battery file.

PR #320, opened 2022-04-04
new: add pomodoro app

thiswillbeyourgithub:

related to #306 

note : some tests were failing because I used the same png as TimerApp so I ended up duplicating the file.

Btw the build error is "region `FLASH_TEXT' overflowed by X bytes"

**features :**
* can use any number and duration of timers without constraints (appart from screen size to type it...) 
* can set preset times and cycle through them by swiping horizontaly
* swipe verticaly to change vibration number (i.e. change number of time it vibrates a the end of each pomodoro)
* randomize vibration pulse if vibrating more than 4 times each pomodoro


daniel-thompson:
Simplest way to avoid the failing check would be to move the app to the apps/ directory (e.g. enable it by default only on the simulator) and then, if you want it on the watch, you can add it to the manifest in your own checkout.

thiswillbeyourgithub:
Done :)

fgaz (commented):
There are extraneous changes in the morse app. They should probably be removed or moved to a separate pr

thiswillbeyourgithub:
I broke my commit tree and can't seem to get out of it cleanly so I'm closing this PR and creating a new one shortly, sorry :/

PR #325, opened 2022-04-11
new: brightness and color change of flashlight

thiswillbeyourgithub:
Touch the flashlight app to cycle through brightness levels as well as colors (I find red light more agreable when I'm lost at night but often need the white light too).

### DIFF
--- a/wasp/apps/flashlight.py
+++ b/wasp/apps/flashlight.py
@@ -4,14 +4,13 @@
 """Flashlight
 ~~~~~~~~~~~~~
 
-Shows a pure white screen with the backlight set to maximum.
+Shows a bright screen that you can tap to change brightness or switch to redlight.
 
 .. figure:: res/TorchApp.png
     :width: 179
 """
 
 import wasp
-
 import icons
 
 class TorchApp(object):
@@ -21,14 +20,16 @@ class TorchApp(object):
 
     def foreground(self):
         """Activate the application."""
-        self.draw()
         wasp.system.request_tick(1000)
+        wasp.system.request_event(wasp.EventMask.TOUCH)
 
         self._brightness = wasp.system.brightness
         wasp.system.brightness = 3
+        self.n_touch = 0
+        self.draw()
 
     def background(self):
-        """De-activate the application (without losing state)."""
+        """De-activate the application (without losing original state)."""
         wasp.system.brightness = self._brightness
 
     def tick(self, ticks):
@@ -36,4 +37,14 @@ class TorchApp(object):
 
     def draw(self):
         """Redraw the display from scratch."""
-        wasp.watch.drawable.fill(0xffff)
+        if self.n_touch % 6 < 3:
+            wasp.watch.drawable.fill(0xffff)
+        else:
+            wasp.watch.drawable.fill(0xf800)
+        print("ok")
+
+    def touch(self, event):
+        self.n_touch += 1
+        n = (wasp.system.brightness - 1) % 3
+        wasp.system.brightness = n if n else 3
+        self.draw()
--- a/wasp/apps/flashlight.py
+++ b/wasp/apps/flashlight.py
@@ -13,6 +13,7 @@ Shows a bright screen that you can tap to change brightness or switch to redligh
 import wasp
 import icons
 
+
 class TorchApp(object):
     """Trivial flashlight application."""
     NAME = 'Torch'
@@ -20,13 +21,13 @@ class TorchApp(object):
 
     def foreground(self):
         """Activate the application."""
-        wasp.system.request_tick(1000)
+        wasp.system.request_tick(3000)
         wasp.system.request_event(wasp.EventMask.TOUCH)
 
         self._brightness = wasp.system.brightness
         wasp.system.brightness = 3
-        self.n_touch = 0
-        self.draw()
+        self._ntouch = 1
+        wasp.watch.drawable.fill(0xffff)  # white
 
     def background(self):
         """De-activate the application (without losing original state)."""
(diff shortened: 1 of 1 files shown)


wasp/apps/flashlight.py
```diff
@@ -4,14 +4,13 @@
...
 
-Shows a pure white screen with the backlight set to maximum.
+Shows a bright screen that you can tap to change brightness or switch to redlight.
 
 .. figure:: res/TorchApp.png
     :width: 179
 """
 
 import wasp
-
```
fgaz, line 14: you left a stray whitespace change here

wasp/apps/flashlight.py
```diff
@@ -21,19 +20,31 @@ class TorchApp(object):
...
         wasp.system.keep_awake()
 
     def draw(self):
         """Redraw the display from scratch."""
-        wasp.watch.drawable.fill(0xffff)
+        if self.n_touch % 6 < 3:
+            wasp.watch.drawable.fill(0xffff)
+        else:
+            wasp.watch.drawable.fill(0xf800)
+        print("ok")
```
fgaz, line 44: the print statement should probably be removed too

wasp/apps/flashlight.py
```diff
@@ -21,19 +20,31 @@ class TorchApp(object):
 
     def foreground(self):
         """Activate the application."""
-        self.draw()
         wasp.system.request_tick(1000)
+        wasp.system.request_event(wasp.EventMask.TOUCH)
 
         self._brightness = wasp.system.brightness
         wasp.system.brightness = 3
+        self.n_touch = 0
```
fgaz, line 28: private attributes should start with `_` (like `_brightness`). besides, shouldn't this line go in `__init__`?

wasp/apps/flashlight.py
```diff
@@ -21,19 +20,31 @@ class TorchApp(object):
 
     def foreground(self):
         """Activate the application."""
-        self.draw()
         wasp.system.request_tick(1000)
+        wasp.system.request_event(wasp.EventMask.TOUCH)
 
         self._brightness = wasp.system.brightness
         wasp.system.brightness = 3
```
fgaz, line 27: I think you need to restore the brightness from the last time the app was opened here

wasp/apps/flashlight.py
```diff
@@ -21,19 +20,31 @@ class TorchApp(object):
...
     def tick(self, ticks):
         wasp.system.keep_awake()
 
     def draw(self):
         """Redraw the display from scratch."""
-        wasp.watch.drawable.fill(0xffff)
+        if self.n_touch % 6 < 3:
+            wasp.watch.drawable.fill(0xffff)
+        else:
+            wasp.watch.drawable.fill(0xf800)
```
fgaz, line 43: Ideally we should skip redrawing if the previous colour was the same (ie. only redraw at `n_touch % 3 == 0`)

wasp/apps/flashlight.py
```diff
@@ -21,19 +20,31 @@ class TorchApp(object):
...
+        print("ok")
+
+    def touch(self, event):
+        self.n_touch += 1
```
fgaz, line 47: _technically_ this will overflow at some point, even though in practice it would take far too many taps :)

```suggestion
        self.n_touch = (self.n_touch + 1) % 6
```

wasp/apps/flashlight.py
```diff
@@ -21,19 +20,31 @@ class TorchApp(object):
...
+        n = (wasp.system.brightness - 1) % 3
+        wasp.system.brightness = n if n else 3
```
fgaz, line 49: why the if? aren't you already changing the brightness by 1?

wasp/apps/flashlight.py
thiswillbeyourgithub: thanks

wasp/apps/flashlight.py
thiswillbeyourgithub: thanks

wasp/apps/flashlight.py
thiswillbeyourgithub: thanks for the "_". There is currently no __init__ implemented so I didn't know if it was worth adding it as I'm not familiar with how to optimize compiled code. Do you think I should add it? The __init__ method would be almost empty.

wasp/apps/flashlight.py
thiswillbeyourgithub: I thought it was best to make the code as small as possible instead of as power saving as possible.

wasp/apps/flashlight.py
thiswillbeyourgithub: The issue is that when it goes to 0 the watch screen turns off and the watch goes back to home. I couldn't find a clever way to code it :/

wasp/apps/flashlight.py
fgaz: It depends! In this case I think adding that code is worth it, since redrawing takes a lot of time and in the meantime the watch will be unresponsive

wasp/apps/flashlight.py
fgaz: > Do you think I should add it?

If the behaviour we want is

> to have the flashlight always start at the same point

then you're right in putting it here in foreground()

wasp/apps/flashlight.py
fgaz: This should do it:

```python
wasp.system.brightness = (wasp.system.brightness % 3) + 1
```

fgaz: oops, it goes the wrong way. you want it to decrease, right?

thiswillbeyourgithub: I slightly modified mine in 12df35b, thanks

wasp/apps/flashlight.py
thiswillbeyourgithub: I don't know if redrawing "red" over "red" is automatically skipped but currently it's super fast on my watch.

I'll add your fix tho

thiswillbeyourgithub: Done. Actually I ended up removing draw() entirely to make it more compact.

I also request tick every 3 seconds instead of 1.

36a267b

Issue #326, opened 2022-04-13
time.mktime() on simulator does not take daylight saving time into account


thiswillbeyourgithub:
Hi,

Just noticed a discrepancy between pinetime and simulator.

I think that `wasp.watch.time.mktime(wasp.watch.rtc.get_localtime())` returns the correct value on the watch but the simulator returns an offset of 1h.

I'm living in france and we recently changed daylight time and I had not noticed this error before, maybe it's due to that.

Issue #298, opened 2022-01-26
inconsistency between simulator and pinetime


thiswillbeyourgithub:
![image](https://user-images.githubusercontent.com/26625900/151224981-97c28375-23f7-4c93-b9a6-0ec9e2550598.png)

as you can see in the picture, running `time.mktime(watch.rtc.get_localtime())` on the watch (via console) returns the unix time.

On the simulator, `watch.rtc.get_localtime()` [calls](https://github.com/daniel-thompson/wasp-os/blob/01c173927ec70467b74652297518bf7e2032cfaf/wasp/boards/simulator/watch.py#L120) `time.localtime()[:8]`.

The result is error " illegal time tuple argument"

PR #327, opened 2022-04-13
fix: inconsistency between simulator time and watch time

thiswillbeyourgithub:
PR to fix isue #298
NOT related apparently to #326

### DIFF
--- a/wasp/boards/simulator/watch.py
+++ b/wasp/boards/simulator/watch.py
@@ -117,7 +117,7 @@ class RTC(object):
         #if self.uptime < 60:
         #    # Jump back a little over a day
         #    return time.localtime(time.time() - 100000)
-        return time.localtime()[:8]
+        return time.localtime()
 
     def get_time(self):
         now = self.get_localtime()


daniel-thompson:
Isn't this something of a see-saw change that trades on inconsistancy for another?

Once this change happens then watch.rtc.get_localtime() will return different length tuples on the watch and on the simulator (so apps that use the ninth value will work on simulator but not on real watch).

daniel-thompson:
PS one way tackle this without the see-saw effect would be to monkey patch time.mktime() on the simulator so I will accept a tuple with only 8 elements.

thiswillbeyourgithub:
I'm afraid this is way above my head so I think I won't be able to provide anything other then having noticed this issue.

daniel-thompson:
> I'm afraid this is way above my head so I think I won't be able to provide anything other then having noticed this issue.

Monkey patching is jargon... it just means changing the implementation of time.mktime by dropping in something new. A bit like:

~~~
import time

py_mktime = time.mktime

def upy_mktime(t):
    if len(t) == 8:
        t = tuple(t) + (-1,)
    return py_mktime(t)

time.mktime = upy_mktime

print(time.mktime(time.localtime()))
print(time.mktime((2022, 8, 16, 20, 25, 52, 1, 228, -1)))
print(time.mktime((2022, 8, 16, 20, 25, 52, 1, 228)))
~~~

I think adding something like this to the simulator board files would make it possible to call time.mktime() with only 8 arguments.

PR #336, opened 2022-06-14
Add a Raise To Wake feature

aeastw:
This commit adds a setting (off by default) that allows the watch
to wake up when the user raises it to look at the screen. This
functionality has mostly been translated from the way that infinitime
implements the feature.

Tested and developed on the Pinetime.

Signed-off-by: Ashley Eastwood <aeastw.git@fastmail.co.uk>


aeastw:
Hi, this is my first attempt at contributing to wasp-os, feedback greatly appreciated. I have tested this on a Pinetime, but don't have other watches available. I suspect there will be some battery impact when the 'Raise To Wake' feature is enabled, I'm looking at testing this over the next few days. Thanks!

PeterCxy:
Might be worth considering implementing this instead: https://github.com/InfiniTimeOrg/InfiniTime/pull/826

FintasticMan:
I don't know much about Wasp-os, so I'm not sure of the specifics. I just wanted to say that both the current implementation of raise to wake in InfiniTime and my implementation rely very heavily on the fact that the motion values are updated every 100 ms, so at 10 Hz.

From looking at this code briefly, it looks like that would mean that the _do_raise_wake method would need to be called at 10 Hz. If this isn't the case in Wasp-os, the values will need to be changed to accommodate that.

aeastw:
Hi @FintasticMan, thanks for the input, and the initial implementation in infinitime! The _do_raise_wake method is called as close to every 100ms as possible (Line 502 has the check to see if this time has expired).
I can see how your new and improved method that's currently in PR relies heavily on the frequency (as it is taking into account speed), but the current implementation of raise to wake (which this PR is based off) doesn't seem to be particularly reliant on the update frequency. Unlike 'shake to wake' which also relies on speed.

aeastw:
@PeterCxy great news, I can update this branch to use the newer implementation and see where we get to once that's done. Thanks for trying it out.

PeterCxy:
@aeastw Never mind, it actually has quite a few issues (referring to my implementation). I'm trying to debug for the reason why, but for now the original algorithm should be the best bet :)

PeterCxy:
Turns out the port of the new impl only did not work due to my own stupid mistake, fixed in <https://gitea.angry.im/PeterCxy/wasp-os/commit/04380f9362b816245def86184060b7363b6b319d>.

FintasticMan:
I haven't tried it out using the built-in tilt wrist interrupt, but [this](https://github.com/InfiniTimeOrg/InfiniTime/blob/373289c072b701684520fce72801136c58b61c55/src/drivers/Bma421.cpp#L118) seems to suggest that the motion sensor is mounted in the PineTime such that the x and y axes are swapped. This would probably mean that the tilt wrist interrupt would interrupt on movement along the x axis, rather than the y axis.

PeterCxy:
@FintasticMan Seems like it is possible to remap the axes on the hardware side from the data sheet. Specifically for wasp-os, there is support for hardware-based axes remapping in [the library] (https://github.com/daniel-thompson/bma42x-upy/blob/b968b2530440afae8b555613e89b1240afcf4da2/BMA42X-Sensor-API/bma42x.h#L541) used by wasp-os. I'm not sure why InfiniTime did not make use of this hardware feature, or maybe someone tried and it did not work?

PeterCxy:
Page 37 in the BMA425 datasheet describes how you can set the axes remap registers for Tilt on Wrist interrupts.

fgaz:
> I'm not sure why InfiniTime did not make use of this hardware feature

in #275 I didn't know that was possible so I did the remapping in software. It'd be nice to use the hardware feature instead!

FintasticMan:
I've found out why InfiniTime doesn't use the wrist tilt interrupt: the BMA421 accelerometer used on older PineTimes doesn't support it.

PeterCxy:
At least in bma42x-upy, the BMA421 implementation includes the same feature configs as BMA425, however I cannot find an official detailed datasheet of BMA421 and cannot confirm whether it can actually do that...

PeterCxy:
Ok according to the discussion [here](https://forum.pine64.org/showthread.php?tid=9617) tilt detection doesn't work on BMA421. But we can probably just implement a fallback for BMA421 and use the hw feature on BMA425.

FintasticMan:
This issue states that it doesn't work: <https://github.com/ck-telecom/pinetime/issues/2>.

PeterCxy:
Yes it probably doesn't work -- but I still think using it on BMA425 and having a generic sw fallback implementation would be nice. I'll take a look soon (TM).

PeterCxy:
@fgaz I've tested the hardware remapping and apparently it only works for "interrupt features" (i.e. wrist twist detection etc) and not for raw data. So we'll need both sw and hw remapping anyway.

FintasticMan:
Would it have an effect on the step tracking then? A little bit strange that it is reasonably accurate while having 2 axes switched. Maybe it just checks for acceleration along any axis though.

PeterCxy:
As an update: I was able to get the interrupt working, but the wrist tilt detection is inconsistent at best and not really working as I expected. The double-tap detection does work quite well though.

@FintasticMan Accoridng to the data sheet, step tracking is also part of the "features" affected by axes remapping.

PeterCxy:
Turns out the reason why hardware gesture detection did not work was due to the mapping -- we need to swap x, y and flip *all* 3 axes to match the coordinate system used by the hardware algorithms. I now have a working prototype and will be cleaning it up and making a new PR some time soon.

PeterCxy:
Unfortunately, implementing the hardware-based gesture detection plus software fallback on the master branch makes the image size just around 1k too big to fit. Maybe some of the apps in wasp/apps should really be moved outside and not included by default...

kLabz:
Could those be two different builds? (at least 2 pinetime builds, don't know about the other targets)

PeterCxy:
@kLabz That might only confuse people than anything else. (I've submitted my PR as #340, maybe the discussion about hw gestures can continue there)

FintasticMan:
@aeastw You might want to uodate this pull request for the version of my raise to wake algorithm that was finally merged in InfiniTime. See the 3 most recent commits on InfiniTimeOrg/InfiniTime#826.

PR #337, opened 2022-06-15
Use base64 encoding for binary upload

fgaz:
Binary upload is _way_ faster now!

The old encoding had a 300% overhead
Base64 has a 33% overhead

Signed-off-by: Francesco Gazzetta <fgaz@fgaz.me>


fgaz:
@daniel-thompson ping

PR #340, opened 2022-06-19
Implement hardware-based raise-to-wake gesture with software fallback

PeterCxy:
With the BMA425 accelerometer shipped with newer PineTime units, it is
possible to configure gesture detection on the hardware and send
interrupts to the nRF chips when detected. The supported gestures
include wrist-twist, double-tap and single-tap.

This commit implements optional raise-to-wake support based on such a
hardware feature. However, for compatibility with earlier units, which
has a BMA421 sensor that does not support gesture detection, we also
introduce a software-based detection algorithm for fallback. This
algorithm is based on the InfiniTime equivalent
<https://github.com/InfiniTimeOrg/InfiniTime/pull/826>.

Due to the way the accelerometer is mounted inside the PineTime, axes
remapping is needed for the hardware to detect the gestures correctly.
This was already implemented in the bma42x driver, and only needs to be
exposed to the Python side via bma42x-upy, which will have a
corresponding pull request for that.

Note that the hardware will never remap the raw data readout -- the
remapping information provided to the hardware will only be used in the
hardware-based algorithms, such as step counting and gesture detection.

Previously in #275, software-based axes remapping was already
introduced, however the mapping was inconsistent with what the hardware
expects. The hardware expects a left-handed coordinate system flipped
around the origin: when worn on the left wrist, the directions down, left,
and towards the user should be positive. On the PineTime, this translates
to all axes being flipped *and* the x and y axes swapped.

The software mapping is now corrected to agree with what the hardware
expects. This also means that apps making use of the accelerometer (for
now, just the Level app) will need to be changed, but that is beyond the
scope of this commit.

Co-authored-by: Ashley Eastwood <aeastw.git@fastmail.co.uk>
Co-authored-by: Finlay Davidson <finlay.davidson@coderclass.nl>

### DIFF
--- a/wasp/apps/settings.py
+++ b/wasp/apps/settings.py
@@ -39,7 +39,8 @@ class SettingsApp():
         self._yy = wasp.widgets.Spinner(160, 60, 20, 60, 2)
         self._units = ['Metric', 'Imperial']
         self._units_toggle = wasp.widgets.Button(32, 90, 176, 48, "Change")
-        self._settings = ['Brightness', 'Notification Level', 'Time', 'Date', 'Units']
+        self._raise_to_wake_toggle = wasp.widgets.Button(32, 90, 176, 48, "On/Off")
+        self._settings = ['Brightness', 'Notification Level', 'Time', 'Date', 'Units', 'Raise To Wake']
         self._sett_index = 0
         self._current_setting = self._settings[0]
 
@@ -73,6 +74,9 @@ class SettingsApp():
         elif self._current_setting == 'Units':
             if self._units_toggle.touch(event):
                 wasp.system.units = self._units[(self._units.index(wasp.system.units) + 1) % len(self._units)]
+        elif self._current_setting == "Raise To Wake":
+            if self._raise_to_wake_toggle.touch(event):
+                wasp.system.raise_wake = not wasp.system.raise_wake
         self._update()
 
     def swipe(self, event):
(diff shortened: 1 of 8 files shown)


PeterCxy:
Note: this currently does make the image size overflow -- if anyone can give some suggestion on what to do to shrink the size, I'm all ears. My suggestion would be to just remove some of the built-in games and apps and instead make them optional. For now all my attempts at fixing the overflowing size are not successful, except building with `DEBUG=0`, but this would also require `CFLAGS_EXTRA=-Wno-error=incompatible-pointer-types`, as some of the modules use incompatible string types with MicroPython.

PeterCxy:
What I ended up doing on my own personal builds is this: <https://gitea.angry.im/PeterCxy/wasp-os/commit/40c3928ce1824fcb2ce7963cb0767cd000696196>. Not sure what `DEBUG=0` does, but all the debugging info on the Python side seems to still be there. Of course, on my personal builds I also have a bunch of other apps removed, but having `DEBUG=0` allows the master branch to fit within the limit.

PeterCxy:
@fgaz How would the icons be initialized if they are moved to the flash?

fgaz:
#403 solved the size problem. @PeterCxy if you rebase and signoff your commit we can push this forward

FintasticMan:
The software fallback should probably be updated to match the version of my algorithm that was eventually merged into InfiniTime.

PR #344, opened 2022-07-20
feat: save alarms across reboots

thiswillbeyourgithub:
- docs: mention that pressing the button will stop the watch
- feat: save alarms across reboots
- new: increment minutes by 5 instead of 1

edit : so far only tested on the simulator because my watch has a weird non related issue that makes troubleshooting a bit difficult at the moment.

### DIFF
--- a/wasp/apps/alarm.py
+++ b/wasp/apps/alarm.py
@@ -102,6 +102,7 @@ class AlarmApp:
                 self.num_alarms += 1
         except Exception:
             pass
+        self._set_pending_alarms()
 
     def foreground(self):
         """Activate the application."""
--- a/wasp/apps/alarm.py
+++ b/wasp/apps/alarm.py
@@ -89,7 +89,7 @@ class AlarmApp:
 
         self.num_alarms = 0
         try:
-            with open("alarms", "r") as f:
+            with open("alarms.txt", "r") as f:
                 alarms = "".join(f.readlines()).split(";")
             for alarm in alarms:
                 n = self.num_alarms
@@ -142,7 +142,7 @@ class AlarmApp:
         try:
             if self.num_alarms == 0:
                 return
-            with open("alarms", "w") as f:
+            with open("alarms.txt", "w") as f:
                 for n in range(self.num_alarms):
                     al = self.alarms[n]
                     f.write(",".join(map(str, al)) + ";")
--- a/wasp/apps/alarm.py
+++ b/wasp/apps/alarm.py
@@ -6,7 +6,7 @@
 ~~~~~~~~~~~~~~~~~~~~
 
 An application to set a vibration alarm. All settings can be accessed from the Watch UI.
-Press the button to turn of ringing alarms.
+Press the button to turn off ringing alarms.
 
     .. figure:: res/AlarmApp.png
         :width: 179
@@ -91,10 +91,10 @@ class AlarmApp:
         try:
             with open("alarms", "r") as f:
                 alarms = "".join(f.readlines()).split(";")
-                for alarm in alarms:
-                    n = self.num_alarms
-                    self.alarms[n][:] = map(int, alarm.split(","))
-                    self.num_alarms += 1
+            for alarm in alarms:
+                n = self.num_alarms
+                self.alarms[n][:] = map(int, alarm.split(","))
+                self.num_alarms += 1
         except Exception:
             pass
 
--- a/wasp/apps/alarm.py
+++ b/wasp/apps/alarm.py
@@ -140,6 +140,8 @@ class AlarmApp:
 
         self._set_pending_alarms()
         try:
+            if self.num_alarms == 0:
+                return
             with open("alarms", "w") as f:
                 for n in range(self.num_alarms):
                     al = self.alarms[n]
--- a/wasp/apps/alarm.py
+++ b/wasp/apps/alarm.py
@@ -103,7 +103,7 @@ class AlarmApp:
 
         self.del_alarm_btn = widgets.Button(170, 204, 70, 35, 'DEL')
         self.hours_wid = widgets.Spinner(50, 30, 0, 23, 2)
-        self.min_wid = widgets.Spinner(130, 30, 0, 59, 2)
+        self.min_wid = widgets.Spinner(130, 30, 0, 59, 2, 5)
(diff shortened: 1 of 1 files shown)


wasp/apps/alarm.py
```diff
@@ -6,6 +6,7 @@
 ~~~~~~~~~~~~~~~~~~~~
 
 An application to set a vibration alarm. All settings can be accessed from the Watch UI.
+Press the button to turn of ringing alarms.
```
daniel-thompson, line 10: s/of/off/

wasp/apps/alarm.py
```diff
@@ -87,14 +87,16 @@ def __init__(self):
...
-            alarm[_HOUR_IDX] = 8
-            alarm[_MIN_IDX] = 0
-            alarm[_ENABLED_IDX] = 0
-        self.alarms[0][_ENABLED_IDX] = _WEEKDAYS
-
+        self.num_alarms = 0
+        try:
+            with open("alarms", "r") as f:
+                alarms = "".join(f.readlines()).split(";")
+                for alarm in alarms:
```
daniel-thompson, line 94: for loop should be outdented one level (it does not use f).

thiswillbeyourgithub:
A bit verbose but it works fine on my pinetime. Any suggestions?

thiswillbeyourgithub:
It seems to sometimes skip the alarm and I don't know why. But at least it loads them fine at boot.

wasp/apps/alarm.py
```diff
@@ -86,21 +87,29 @@ def __init__(self):
...
+            pass
+        self._set_pending_alarms()
 
     def foreground(self):
         """Activate the application."""
 
         self.del_alarm_btn = widgets.Button(170, 204, 70, 35, 'DEL')
         self.hours_wid = widgets.Spinner(50, 30, 0, 23, 2)
-        self.min_wid = widgets.Spinner(130, 30, 0, 59, 2)
+        self.min_wid = widgets.Spinner(130, 30, 0, 59, 2, 5)
```
fgaz, line 112: Subjective I know, but I liked the minute precision...

wasp/apps/alarm.py
```diff
@@ -136,6 +145,16 @@ def background(self):
...
         self._set_pending_alarms()
+        try:
+            if self.num_alarms == 0:
+                return
+            with open("alarms.txt", "w") as f:
+                for n in range(self.num_alarms):
+                    al = self.alarms[n]
+                    f.write(",".join(map(str, al)) + ";")
+        except Exception:
+            pass
```
fgaz, line 156: why are exceptions ignored here? shouldn't the user be aware of any write failures?

fgaz (commented):
Late review, sorry

wasp/apps/alarm.py
```diff
@@ -86,21 +87,29 @@ def __init__(self):
...
+            with open("alarms.txt", "r") as f:
+                alarms = f.readlines()[0].split(";")
+            if "" in alarms:
+                alarms.remove("")
+            for alarm in alarms:
+                n = self.num_alarms
+                h, m, st = map(int, alarm.split(","))
+                self.alarms[n][0] = h
+                self.alarms[n][1] = m
+                self.alarms[n][2] = st
```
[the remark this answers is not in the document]
thiswillbeyourgithub: You mean instead of 0 1 and 2?

wasp/apps/alarm.py
```diff
@@ -86,21 +87,29 @@ def __init__(self):
...
+                self.num_alarms += 1
+        except Exception:
+            pass
```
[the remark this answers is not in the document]
thiswillbeyourgithub: I originaly removed the defaults because it overwrote the alarms in case of read/write failure when I was prototyping.

I ended up leaving it removed because I don't think it's a particularly good idea to have a default alarm. Your call :)

wasp/apps/alarm.py
thiswillbeyourgithub: I recently had very annoying IO failures because I didn't have space left because of my sleep logging. I ended up leaving it like that but am still all ears.

wasp/apps/alarm.py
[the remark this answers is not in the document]
fgaz: Yes

wasp/apps/alarm.py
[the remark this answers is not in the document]
fgaz: I don't have any strong opinion about this, I think it's fine either way. Just thought you forgot to add it back

wasp/apps/alarm.py
fgaz: I think in that case the user should be alerted. If I expect the alarm to be saved and someday it silently doesn't happen I could miss an alarm

wasp/apps/alarm.py
fgaz: While it's true that precise control over the alarm is mostly useful with the timer app, sometimes I want to be alerted on absolute --not relative-- time, and except for wake up alarms a 5 min precision is a bit rough. Imagine setting an alarm 5 mins before the bus arrives for example (so you can go to the stop). With 5 mins precision the interval between the alarm and when the bus arrives can go up to 9 mins.

Ideally I think we could have a way for the user to control the speed of spinners, like a >> button above the > button, or detection of fast taps or long presses, like most alarm clocks. But that's a bigger feature request.

wasp/apps/alarm.py
fgaz: For example by simply opening a PagerApp with a message like "Warning: unable to save your alarm to flash storage, it will not persist a reboot."

wasp/apps/alarm.py
thiswillbeyourgithub: I think it would be better to add long press support : https://github.com/daniel-thompson/wasp-os/issues/331

wasp/apps/alarm.py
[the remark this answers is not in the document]
thiswillbeyourgithub: Thanks!
See  https://github.com/daniel-thompson/wasp-os/pull/350

wasp/apps/alarm.py
thiswillbeyourgithub: Thanks!
See  https://github.com/daniel-thompson/wasp-os/pull/350

PR #350, opened 2022-08-17
feat saved alarms 2

thiswillbeyourgithub:
- quick fix: using indices for alarms
- fix: showing exception if saving alarms failed

### DIFF
--- a/wasp/apps/alarm.py
+++ b/wasp/apps/alarm.py
@@ -20,6 +20,7 @@ import time
 import widgets
 import array
 from micropython import const
+from apps.pager import PagerApp
 
 # 2-bit RLE, generated from res/alarm_icon.png, 390 bytes
 icon = (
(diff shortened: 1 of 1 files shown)


wasp/apps/alarm.py
```diff
@@ -96,12 +97,13 @@ def __init__(self):
...
-                self.alarms[n][2] = st
+                self.alarms[n][_HOUR_IDX] = h
+                self.alarms[n][_MIN_IDX] = m
+                self.alarms[n][_ENABLED_IDX] = st
                 self.num_alarms += 1
-        except Exception:
-            pass
+        except Exception as err:
+            wasp.system.switch(PagerApp("Error when reloading alarms: "
+                                         "'{}'".format(err)))
```
fgaz, line 106: Will this trigger also the first time the app is run, when the file is not there yet?

thiswillbeyourgithub: Yes. My personnal inclination is of course to pay attention to this kind of things but on waspos given that every line seems to count I felt that was a good compromise.

What do you think?

PR #353, opened 2022-08-19
new: import step counter later

thiswillbeyourgithub:
When disabling the stepcounter in the manifest, the watch refuses to boot without issuing error. I think this fix would make it more clear and easier to save on RAM.


fgaz:
For this to be useful, the step counter should be made optional (through wasp.toml), but currently it is always included

PR #361, opened 2022-09-20
add comment warning to ignore IDE asking to remove unused imports

thiswillbeyourgithub:
Hi,

Very simple PR to add a comment for newcomers to ignore the PEP8 warning to remove "import Pin" as it stops the battery level from updating as it was one of the first thing I did...


daniel-thompson:
Change looks OK... but the checks are not passing!

thiswillbeyourgithub:
Thanks, done!

PR #362, opened 2022-09-21
Minor: move Stopwatch library from widgets.py to apps/stopwatch.py

thiswillbeyourgithub:
Isn't it tidier?

### DIFF
--- a/wasp/apps/stopwatch.py
+++ b/wasp/apps/stopwatch.py
@@ -21,7 +21,7 @@ class StopwatchApp():
     ICON = icons.app
 
     def __init__(self):
-        self._timer = wasp.widgets.Stopwatch(120-36)
+        self._timer = Stopwatch(120-36)
         self._reset()
 
     def foreground(self):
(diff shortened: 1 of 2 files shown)


kLabz:
It's also used in [wasp/apps/sports.py](https://github.com/daniel-thompson/wasp-os/blob/e169dbf61a870b594bb9fe52f02fa1e002bf9373/wasp/apps/sports.py#L22) though

thiswillbeyourgithub:
In commit 18aa61 I also enabled hours/Minute conversion instead of just second display.

@kLabz omg I feel dumb. Well I'm closing this then. Thanks!

thiswillbeyourgithub:
Come to think of it, isn't it an indicator that having 2 apps for this is too much? I never used the sports app and am curious if people really find useful to see both the stopwatch and the step counter

daniel-thompson:
> Come to think of it, isn't it an indicator that having 2 apps for this is too much? I never used the sports app and am curious if people really find useful to see both the stopwatch and the step counter

The sports app doesn't show the daily step count, it starts from zero. The idea is you can set the Sports app running at the start of a run and when you get to the end you know how many strides and how long it took.

thiswillbeyourgithub:
That seems a bit specific too me and I would think it would be better to leave it in /apps instead of wasp/apps. Your call as usual.

PR #366, opened 2022-09-25
Adding fonts.sans24scale

drtonyr:
Signed-off-by: Tony Robinson <tonyr@tonyRobinson.com>

### DIFF
--- a/wasp/fonts/__init__.py
+++ b/wasp/fonts/__init__.py
@@ -5,6 +5,7 @@ import fonts.sans18 as sans18
 import fonts.sans24 as sans24
 import fonts.sans28 as sans28
 import fonts.sans36 as sans36
+import fonts.sans24scale as sans24scale
 
 def height(font):
     return font.height()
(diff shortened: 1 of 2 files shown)


daniel-thompson:
Scaling is a great feature.

However I think it might be better implemented in draw565.py instead. Vertical scaling is almost "free" if implemented in _draw_glyph() since we can just send the same pixels to the hardware twice. Similar there is plenty of opportunity for a fast horizontal pixel doubling version of _bitblit().

If implemented in draw565.py it would also be useful to update the TestApp to benchmark pixel doubling.

drtonyr:
Thanks - I'm glad you like the idea.

I had a choice, a minimal touch change that could easily be adopted and maintained all current interfaces, or a more extensive change that would be more efficient.   I chose an extreme version of the former (at the moment nothing is enabled by default) because this was my first pull request, I didn't want to mess anything up and I didn't know if the blocky nature of scaling would be acceptable.  However, as things stand the dynamically generated fonts end up in RAM which is very scarce, there is some horrible bit unpacking/packing that could/should be ditched and it's all quite slow (even with the base n hacks).

Let's not take this route forward.  I would like to understand draw565.py much better.   I have a few pull requests and some dev work to get out for discussion before looking at draw565.py.

PR #368, opened 2022-09-29
feat: persistent data storage across reboots for apps

thiswillbeyourgithub:
This creates 2 new methods for wasp.system to store and get settings. They will be stored in files in a settings folder.

I'm testing this with stopwatch (start time) and alarms.

What do you think?

### DIFF
--- a/wasp/wasp.py
+++ b/wasp/wasp.py
@@ -21,6 +21,7 @@ import steplogger
 import sys
 import watch
 import widgets
+import os
 
 from apps.launcher import LauncherApp
 from apps.pager import PagerApp, CrashApp, NotificationApp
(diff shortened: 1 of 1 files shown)


thiswillbeyourgithub:
Btw I initially wanted to use asset statements when saving data but apparently they were working on the simulator but ignored on the pinetime.

Reverting the assert in [this commit][https://github.com/thiswillbeyourgithub/wasp-os/commit/86dc51535435351d37be5f636afeac9ff80f8867) fixed the issue.

Is it normal? Is there something I'm missing? Can we improve the simulator consistency?

wasp/wasp.py
```diff
@@ -595,4 +596,41 @@ def theme(self, theme_part: str) -> int:
         idx = theme_parts.index(theme_part) * 2
         return (self._theme[idx] << 8) | self._theme[idx+1]
 
+    def store_settings(self, name, value):
```
daniel-thompson, line 599: store/get don't really pair together. Either get/set or load/store (although to be honest I'd be tempted to go for getopt/setopt since, although it is short I think it conveys enough meaning... although that might be because I'm used to the Unix getopt call).

wasp/wasp.py
```diff
@@ -595,4 +596,41 @@ def theme(self, theme_part: str) -> int:
...
+            return True
+        except Exception as err:
+            self.notify(watch.rtc.get_uptime_ms(), {
+                "src": "Persistent settings",
+                "title": "Storing failed",
+                "body": "Error when storing '{}': '{}'".format(name, err)})
+
+    def get_settings(self, name, delete=False):
+        if "settings" not in os.listdir():
+            os.mkdir("settings")
```
daniel-thompson, line 624: No need to mkdir here, we can just return None/default.

wasp/wasp.py
```diff
@@ -595,4 +596,41 @@ def theme(self, theme_part: str) -> int:
...
+                    f.write(str(value))
+            assert name in os.listdir("settings/")
+            return True
+        except Exception as err:
+            self.notify(watch.rtc.get_uptime_ms(), {
+                "src": "Persistent settings",
+                "title": "Storing failed",
+                "body": "Error when storing '{}': '{}'".format(name, err)})
+
+    def get_settings(self, name, delete=False):
```
daniel-thompson, line 622: Adding default = None might make it better for some callers (and then return default rather than None in the error paths).

wasp/wasp.py
```diff
@@ -595,4 +596,41 @@ def theme(self, theme_part: str) -> int:
...
+        if "settings" not in os.listdir():
+            os.mkdir("settings")
+        if name not in os.listdir("settings/"):
+            return None
```
daniel-thompson, line 626: Do we need pre-checks? We could just use try/except (and return None/default on the except path)

daniel-thompson (changes requested):
Overall I like this... just a few small suggestions

wasp/wasp.py
thiswillbeyourgithub: Agreed. I went to simply "get/set" in the latest push.

wasp/wasp.py
thiswillbeyourgithub: Agreed, removed in latest push.

wasp/wasp.py
thiswillbeyourgithub: Added so I added try blocks. I also added a new "return None" to "set" in caes it fails.

wasp/wasp.py
thiswillbeyourgithub: I actually don't understand what you mean by "default" or how people could use it.

thiswillbeyourgithub:
By the wat I made an enhancement to the alarm app to support more than 4 alarms but I'm waiting for this PR to be approved to avoid having too many incompatible PR waiting to be merged :)

PR #369, opened 2022-10-01
Adding wasp-os/apps/fourInARow.py

drtonyr:
As per discussion at https://github.com/daniel-thompson/wasp-os/issues/322

Signed-off-by: Tony Robinson <tony@tonyRobinson.com>


daniel-thompson:
Hi @drtonyr 

The auto tests are currently failing for this PR. In particular there is a check to make sure that all apps have a screenshot (set up a good looking game and press 's' on the simulator) and are documented.

IIRC you can run the checks locally with `make check`.

drtonyr:
> Hi @drtonyr
> 
> The auto tests are currently failing for this PR. In particular there is a check to make sure that all apps have a screenshot (set up a good looking game and press 's' on the simulator) and are documented.
> 
> IIRC you can run the checks locally with `make check`.

I saw the build fail last night, thanks, and that pointed me to 's' for screenshot which I've done.   I didn't know about `make check` - that's a very cool feature (is it documented?).    I'm looking at the rest now.

drtonyr:
Checks are now passing.  In general, I find newbie comments really useful, they are good for things to add to a TO DO list -and I'm a Newbie!

So here are some of the things I've found.

* I find the name mangling really confusing.   On apps.rst I'm now using apps.fourInARow, elsewhere I have '4 ina Row', '4inarow', 'Four In A Row' and 'FourInARow' - and then a lot of these have App added just to increase the combinations.  As earlier, it would be good to settle on whether file name for apps start with an upper char or not.  As a newbie you can waste days on getting this working - I did - it's a real barrier to entry.
* pressing s on the screenshot (as instructed) gave me width 358 but `make check` wants 179 - there could be one key press for full size, one for what is needed.
* Are apps.rst and README.rst autogenerated?  It looks like it is but I edited by hand to get the tests to pass
* There could be a TO DO list so that your small army of helpers could pick up stuff like the above

daniel-thompson:
> * `make check` is essential for a Newbie, but it's not documented

True, it would be good to get that documented in the contributors guide

> * I find the name mangling really confusing.   On apps.rst I'm now using apps.fourInARow, elsewhere I have '4 ina Row', '4inarow', 'Four In A Row' and 'FourInARow' - and then a lot of these have App added just to increase the combinations.  As earlier, it would be good to settle on whether file name for apps start with an upper char or not.  As a newbie you can waste days on getting this working - I did - it's a real barrier to entry.

Class  names shoud follow PEP-8 (FourInARowApp) and the filenames used in the apps/ directory must be the class name with the App removed. NAME can be whatever you think looks best in the launcher.

However in both cases the names must be short enought to be rendered without drawing off the edge of the screen (that's why there are two names in the first place, the Launcher name might have to be shorter). There might be some holes in `make check` but ideally the length limits can be enforced by the test suite (unfortunately the limit is measures in pixels not characters because the font is variable width).

> * pressing s on the screenshot (as instructed) gave me width 358 but `make check` wants 179 - there could be one key press for full size, one for what is needed.

I think you have misunderstood this. There is no need to resize it, 358 is the correct width (just look at all the others .png files).

The watches are HiDPI devices so if you show a 358 wide screenshot on a FullHD laptop it looks way to big. Hence apps.rst and README.rst request a half-size scaling (e.g. on a FullHD screen the browser will downscale by 2x but on a HiDPI screen the browser will show all the pixels).

Perhaps we need to get `make check` to verify the image dimensions (the most comment problem we get with screenshots is people not finding the 's'creenshot feature and using something ad-hoc (which then doesn't look right in the screenshot grid).

> * Are apps.rst and README.rst autogenerated?  It looks like it is but I edited by hand to get the tests to pass

They are not auto-generated (although they are auto-tested).

> * There could be a TO DO list so that your small army of helpers could pick up stuff like the above

The goal is for the entire TODO list is expected to be *"ensure `make check` passes and if you see any failures then look up the comments that explain how to fix things"*. As above I agree it would be good to get this into the contributor's guide.

There may be a couple missing but there are (or should be) comments in the test suite explaining what the test does (and the failure reports give line numbers to help you find them).

Having said that we could definitely make things simpler by moving all apps to CamelCase so it is easy to move them between apps/ and wasp/apps/ .

drtonyr:
Thanks - I'm getting there.

Yes, I probably picked up the 179 dimension because I had a docstring error and I was fixing that.

IMHO it would be cleaner if the NAME was only used in the launcher and the class name used everywhere else.  I see that my image was created with NAME and not the class name.

I do normally keep my code in wasp-os/wasp/apps and not in wasp-os/apps, that way I know it will run.  Making it easy to move between the two would be a huge plus.   You can tell I'm quite confused about what name is needed where, in fact the last submit fails for that reason (I'll fix it in a minute).

I am very impressed with `make check` and I do like the idea that it tells you what to do to fix things.

beardeddude:
As you may know after the merge of #403 the naming conventions and file placement have changed slightly. See https://wasp-os.readthedocs.io/en/latest/appguide.html#app-naming-conventions-and-placement for details.

fgaz:
@drtonyr will you open another pr? the app looked great!

drtonyr:
> @drtonyr will you open another pr? the app looked great!

Ah, I thought this was an amazing coincidence, but now I see that my clean up of my own github space registered here.

Yes, as of a few minutes ago it's running on my watch (it beat me!), so now all I have to do is relearn how to do a proper pull request.

PR #370, opened 2022-10-04
feat: remove any alarm given an action

thiswillbeyourgithub:
I think this is useful. I'm behind an alarm app with snooze and lots of other features so keeping track of every time an action is scheduled can be cumbersome.

### DIFF
--- a/wasp/wasp.py
+++ b/wasp/wasp.py
@@ -351,7 +351,12 @@ class Manager():
     def cancel_alarm(self, time, action):
         """Unqueue an alarm."""
         try:
-            self._alarms.remove((time, action))
+            if time is None:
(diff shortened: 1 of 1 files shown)


thiswillbeyourgithub:
If anyone here has advice on how to make this more efficient I'm all ears! This is needed for SleepTk to work fine.

fgaz:
> If anyone here has advice on how to make this more efficient I'm all ears

I think it's fine. It's O(n), but that list is going to be short anyway, so it's probably better to keep the code short and simple and save memory

fgaz (approved):

daniel-thompson:
Merged. Thanks!

PR #376, opened 2022-10-31
docs: add a short note about the p8 plus variant

leroyshirto:
I recently made the mistake of purchasing a "P8 Plus" instead of a P8. I hope this comment will help others not make a similar mistake.

It sound like this has happened before too, albeit under slightly different circumstances https://github.com/daniel-thompson/wasp-os/issues/259

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -258,6 +258,9 @@ display with touch screen, a step counter and a heart rate sensor.
     boot wasp-os successfully but the step counter application will
     be disabled and cannot function.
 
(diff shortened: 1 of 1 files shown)


fgaz:
@leroyshirto could you sign off your commit? There are detailed instructions [in the details of the failing check](https://github.com/wasp-os/wasp-os/pull/376/checks?check_run_id=9201460721)

fgaz:
Still needs a sign-off

PR #377, opened 2022-11-03
gallery: Fix modulo-by-zero error on swipe

hendursaga:


fgaz:
To fix the failing check you should accept the DCO by using the --signoff flag when committing

fgaz:
Ping :)

hendursaga:
How can I redo the commit, to do the DCO signoff, but also because there are conflicts here, too?

PR #379, opened 2022-11-11
add a test on string width

rcaneill:
This pull request adds a verification on the length of a string before drawing, If the string is too long, it is trimmed.
I made some tests, with long texts and different widths, it seems to work well.

I understand if the mentality should be that programmers verify that their text fits well. Anyway I will keep it for my own fork, it makes it easier to change e.g. the font without breaking everything (if the new font is larger, text will be trimmed, instead of having an app not executing)

### DIFF
--- a/wasp/draw565.py
+++ b/wasp/draw565.py
@@ -310,8 +310,18 @@ class Draw565(object):
         font = self._font
         bg = self._bgfg >> 16
 
+        (w, h) = _bounding_box(s, font)
         if width:
+            wmax = min(width, 240 - x)
+            width = wmax
+        else:
+            wmax = 240 - x
+        while w > wmax:
+            # We need to strip the string
+            s = s[:-1]
             (w, h) = _bounding_box(s, font)
+
+        if width:
             if right:
                 leftpad = width - w
                 rightpad = 0


rcaneill:
(sorry I forgot to sign the commit, I'll try to fix that)

fgaz:
See https://github.com/daniel-thompson/wasp-os/issues/248#issuecomment-917448102

rcaneill:
Yes I had seen this issue, but when I tried to use another font (that was producing larger strings), it was breaking almost every text. That's why I submitted this change.
I in any case am happy to discuss it's utility.

wasp/draw565.py
```diff
@@ -310,8 +310,18 @@ def string(self, s, x, y, width=None, right=False):
         font = self._font
         bg = self._bgfg >> 16
 
+        (w, h) = _bounding_box(s, font)
         if width:
+            wmax = min(width, 240 - x)
+            width = wmax
+        else:
+            wmax = 240 - x
+        while w > wmax:
```
rcaneill, line 319: Maybe it is needed to change to
```
while w > wmax - 1
```
otherwise it does not work if the string has exactly the required width. Best would probably to change the following lines, but I don't have time / I don't understand them fully

drtonyr:
Is there any appetite to continue this work?   I've done a little work with .string() recently so I know the code.   I can also see why this is useful, I have situations where I'd rather the string was truncated than an exception thrown.

if there is appetite, then I have three comments.
1) as already discussed, the 'out by one' in while `w > wmax - 1` needs resolving
2) I think this should this be a truncate=False option.  This would give backward compatibility.
3) I believe that `240` would be better expressed as wasp.watch.display.width

rcaneill:
Hello @drtonyr , I switched to infinitime and don't have time to work on this PR before 6 months or more, so please feel free to take the code as is and continue the PR!

drtonyr:
Thanks @rcaneill for getting back to me so quickly.

I had another thought after I wrote the above.  We have a [.wrap()](https://wasp-os.readthedocs.io/en/latest/wasp.html?highlight=wrap#draw565.Draw565.wrap) method which is related, it takes a string and gives you the breaks so you don't overflow width.

There are several ways we could do this.  I think it's best to close this one here and reopen when we have several use cases as then it'll be easier to pick the best route forward.  Hope you don't mind, reopen if you disagree.

PR #382, opened 2023-01-09
Quick timer enhancement

beardeddude:
Features:
* Added starting screen to timer app with quick timers
* Reduced RAM usage when not opened

TimerApp

TimerApp (copy 1)

I added quick timers to the timer app to make it faster to use. The values I chose should be the most common but I welcome input. If custom is selected the user is presented the same display as before and if a timer is stopped the app will return to the custom set screen with the original time reset (quick or custom) for fast re-use. To get back to the quick select screen the app would need to be closed while the timer is not running.

Testing:
* I only own a Pinetime so testing was done only on it and the simulator.
* Memory usage (ROM) increase seems minimal and RAM use while not open is reduced. RAM usage while open is slightly increased.

Other thoughts:
I have no idea about the current release process but it would be awesome to see a newer release binary on github as the current one is really old.


fgaz:
> I have no idea about the current release process but it would be awesome to see a newer release binary on github as the current one is really old.

I asked on #wasp-os:libera.chat and it looks like this is the blocker: #386

beardeddude:
Conflicts resolved. Ready for review

apps/timer.py
```diff
@@ -88,7 +105,28 @@ def tick(self, ticks):
...
+                        elif btn_idx == _TEN_BUTTON:
+                            self._set_delay(10)
+                        elif btn_idx == _FIFTEEN_BUTTON:
+                            self._set_delay(15)
+                        elif btn_idx == _THIRTY_BUTTON:
+                            self._set_delay(30)
+                        elif btn_idx == _FORTY_FIVE_BUTTON:
+                            self._set_delay(45)
+                        elif btn_idx == _SIXTY_BUTTON:
+                            self._set_delay(60)
```
fgaz, line 127: I wonder if using a dictionary and giving each button an index equal to the time would be more storage/memory-efficient. Something like
```python
_CUSTOM_BUTTON = const(0)
...
self.quick_btns = {
    5: widgets.Button(20, 50, 60, 40, '5'),
    10: widgets.Button(90, 50, 60, 40, '10'),
    ...
    _CUSTOM_BUTTON: widgets.Button(20, 150, 200, 40, 'Custom')
}
...
for btn_idx in keys(self.quick_btns):
    if self.quick_btns[btn_idx].touch(event):
        if btn_idx == _CUSTOM_BUTTON:
            ...
        else:
            self._set_delay(btn_idx)
```

fgaz (commented):
Looks good, but I'm a bit worried about the increase in size. Imo the timer should be a small app that does not prevent other apps from loading.

apps/timer.py
beardeddude: Ya, I made the choice to be as efficient in RAM as I could at cost of ROM. This if else block with constants uses very little RAM when frozen into the firmware. In my experiments the dictionary is pretty RAM heavy. The buttons only exist when the app is opened and are released when it is no longer visible. 

For what its worth this only added a couple KB to my firmware and I have been using it daily for the better part of a year and have not encountered an out of memory error.

PR #383, opened 2023-01-10
Bidirectional notification clearing over gadget bridge

beardeddude:
Currently clearing notifications on the watch leaves the notifications on the phone. Now that Gadgetbridge supports notification clearing we should keep the devices in sync so notification don't need to be cleared twice. 

Features:
* Clear notifications button now clears notifications on device as well

Other:
* Moved send command function from music app to gadget bridge file for reuse between apps

Testing:
* Tested using Pinetime and Gadgetbridge 0.72 (0.56+ should work)


beardeddude:
Conflicts resolved. Ready for review

PR #384, opened 2023-01-11
Phone finder app

beardeddude:
* Added phone finder app that works with Gadgetbridge
* Fixed a bug in ToggleButton
* Moved send_cmd from music player to gadgetbridge.py for all apps to use (same as other PR)

FinderApp

Notes:

* There doesn't appear to be enough room to include this as a system app (despite how small it is) but I wrote the pull request that way anyways. Even removing the icon there is not enough room. I assume something would need to be removed or that this would be added to that optional apps folder. Either way I offer this app with the hope it can be used some how.
* If this should be an optional app in /apps let me know and I''ll update this PR
* The change to mustic player and gadgetbrige are the same as https://github.com/wasp-os/wasp-os/pull/383 and should not conflict with that PR

Testing:
* Was done on a Pinetime with Gadgetbridget 0.72


beardeddude:
Closed because other version was merged

Issue #338, opened 2022-06-16
Adjust battery percentage calculation according to the new acqtime


fgaz:
#254 got closed but only acqtime was fixed. the percentage calculation still has to be adjusted according to the new mV values in the table. See https://github.com/daniel-thompson/wasp-os/issues/254#issuecomment-927263942

Issue #385, opened 2023-01-11
Charging level does not update without rebooting


AeneasChristodoulou:
Hey guys, I'm pretty new to WaspOS. 
My PineTime does not update its charging level in WaspOS without me manually rebooting it.
I'm using the latest "official" version following the installation guide. I'd assume this is just me being stupid here?

beardeddude:
also the official build is very old. If you build yourself you will see lots of improvments

fgaz:
I'm pretty sure it's because of this: https://github.com/wasp-os/wasp-os/pull/318#pullrequestreview-1074764497

Looks like it wasn't reverted yet

PR #387, opened 2023-01-15
fix: battery level is not reset when charging without rebooting

thiswillbeyourgithub:
I noticed the issue of battery level not being reset (#385) and fixed it on my own fork recently. I'm low on time currently but figured I would at least PR my current version.

I'm sorry for all this :/

### DIFF
--- a/wasp/drivers/battery.py
+++ b/wasp/drivers/battery.py
@@ -6,6 +6,7 @@
 """
 import micropython
 from machine import Pin, ADC
+import array
 
 class Battery(object):
     """Generic lithium ion battery driver.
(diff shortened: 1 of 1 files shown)


wasp/drivers/battery.py
```diff
@@ -79,6 +89,5 @@ def level(self):
         :returns: Estimate battery level in percent.
         """
         mv = self.voltage_mv()
-        level = ((19 * mv) // 100) - 660
-        level = min(100, max(0, level))
-        return level
+        level = int((mv - 3500) / (700) * 100)  # 0.7V is 4.2-3.5V
```
fgaz, line 92: Does this fix #338?

thiswillbeyourgithub: Not sure but it assumes the level goes from 3.5 to 4.2V. Obviously the level value will not evolve linearly in that case. Personaly I'm using the voltage value directly to know how much power I have left

PR #392, opened 2023-01-18
Doc update: how to autoload apps in flash

beardeddude:
I'm not sure if this is even good practice. But I learned I could do this by reading the software.py file. I tried to get it to work using the upload command of the wasp tool but couldn't get it to work. But by using the push command I am easily able to store a .mpy so that its automatically added to the software list. 

I personally found this useful but it obviously still has the RAM constraints of a flash based program. I didn't think the docs made it clear this was possible. So I offer this update. If this is encouraging bad practice I'm interested in that as well.

### DIFF
--- a/docs/appguide.rst
+++ b/docs/appguide.rst
@@ -417,6 +417,43 @@ section to add an import and register for you application from ``main.py``
             --exec wasp/drivers/cst816s.py\
             --eval "watch.touch = CST816S(watch.i2c)"`
 
+Auto loading applications in flash
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If there is not enough room on your device to freeze your application
+into the wasp-os binary you can still have it automatically added to
+the software list by uploading it to the apps directory of your device.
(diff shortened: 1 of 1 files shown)


fgaz:
Thanks! This piece of documentation was really needed.

> I'm not sure if this is even good practice.

Yes it is. All apps in `apps/` are made to be uploaded that way.

>  I tried to get it to work using the upload command of the wasp tool but couldn't get it to work.

What was the error? You should use `--binary` because `.mpy` files are not just text.

You should also mention that that the app should follow a naming convention: the file name should be the class name - "App" + ".mpy"/".py", so "ReadMeApp" → "ReadMe.mpy"/"ReadMe.py"

For reference the feature was added in ba8546dd605799c189ccce778b0c468cb129054d

beardeddude:
Thanks for the feedback. I updated the change to use the binary upload (which was what I was missing). And added a section on the naming conventions as I had some trouble figuring them out myself. I think I got them right but sanity check me.

daniel-thompson (approved):

PR #393, opened 2023-01-22
Phone finder app (optional)

beardeddude:
This is the same PR as https://github.com/wasp-os/wasp-os/pull/384 but as an optional app. Putting both up for convenience. I'll delete what isn't used.

FinderApp

* Added phone finder app that works with Gadgetbridge
* Fixed bug in ToggleButton
* Moved send_cmd from music player to gadgetbridge.py for all apps to use (same as https://github.com/wasp-os/wasp-os/pull/383)


tigercoding56:
Ty  (i needed to know how to make a optional app)

daniel-thompson (approved):

PR #400, opened 2023-02-21
User apps_A binary clock face would be cool.

FlyingSquirrel03:
If you could somehow add a simple binary clock face in.... You will be much appreciated.

Example: 
image


fgaz:
This is a pull request, please [open an issue](https://github.com/wasp-os/wasp-os/issues/new/choose) instead

PR #405, opened 2023-03-12
Minesweeper, Therminal watch face, better notifications.

sbot50:
### Added:
- Mines_light as an optional app
- Mines_dark as an optional app
- therm_clock as a watchface
- firacode as a font (size 20, used by therm watchface because monospace looks better)
- editer pager app for better notifications (explained below)

### MinesApp:
Normal minesweeper,
Comes in 2 flavors light mode and the superior dark mode.
Click to reveal a tile, double click to flag or unflag a tile.
(Flagged tiles cant be revealed unless unflagged)

### ThermClockApp:
Therminal watch face taking HEAVY inspiration from the infinitime one.
First watchface with a notification counter instead of just an icon.
FiraCode font needed for monospace because it looks better.

### PagerApp:
Edited NotificationApp so it's actually useful.

**Controls:**
Swipe Left/Right: go from old to new notifications (oldest is left)
Swipe Up: go to the next pages of the notification*
Swipe Up (last page of the notification): Remove notification
Swipe Down: go to the previous pages of the notification*
Swipe Down (first page of the notification): Clear all notifications*
Press Button: go back to watch face (previous application, but thats always watch face in this case)
*Default behaviour that hasn't changed

**Tweaks:**
- Old notifications don't clear after looking at them
- Look at the oldest notification first
- Scroll through notifications
- Clear individual notifications


sbot50:
@daniel-thompson Please accept this, the current notification system is really bad.

PR #406, opened 2023-03-12
Minesweeper, Therminal Watch Face, Better Notification Area

sbot50:
### Added:
- Mines_light as an optional app
- Mines_dark as an optional app
- therm_clock as an optional watchface
- firacode as a font (size 20, used by therm watchface because monospace looks better)
- edited pager app for better notifications (explained below)

### MinesApp:
Normal minesweeper,
Comes in 2 flavors light mode and the superior dark mode.
Click to reveal a tile, double click to flag or unflag a tile.
(Flagged tiles cant be revealed unless unflagged)

### ThermClockApp:
Therminal watch face taking HEAVY inspiration from the infinitime one.
First watchface with a notification counter instead of just an icon.
FiraCode font needed for monospace because it looks better.

### PagerApp:
Edited NotificationApp so it's actually useful.

**Controls:**
Swipe Left/Right: go from old to new notifications (oldest is left)
Swipe Up: go to the next pages of the notification*
Swipe Up (last page of the notification): Remove notification
Swipe Down: go to the previous pages of the notification*
Swipe Down (first page of the notification): Clear all notifications*
Press Button: go back to watch face (previous application, but thats always watch face in this case)
*Default behaviour that hasn't changed

**Tweaks:**
- Old notifications don't clear after looking at them
- Look at the oldest notification first
- Scroll through notifications
- Clear individual notifications

#### SideNote
The firacode font is a free and opensource font for both personal and commercial use.
I added it to the fonts folder but it isn't enabled by default due to it's large size, it's needed for the Therminal watch face however which is why it was added (the only reason its needed is because it was the first working monospace font i could find)
Github: https://github.com/tonsky/FiraCode


sbot50:
for the love of god why does everything need to be signed off

sbot50:
Somehow it ran before but is now awaiting approval?
Anyways, I think it should be reasy for review if the workflows get approved since they pass on my fork

Untitled

sbot50:
Ok finally it should be ready for commit.

It's PEP8 compliant.
The binary and simulation build both succeed on my main branch.
All commits are signed by me.

Sorry for the stupid commit names, but getting the stuff to actually build was kinda a pain

fgaz:
Could you make three separate pull requests, or at least three separate commits for the three features you're adding?

wasp/boards/pinetime/watch.py
```diff
@@ -0,0 +1,136 @@
+# SPDX-License-Identifier: LGPL-3.0-or-later
```
fgaz, line 1: this file should not be committed

sbot50: confused as to how that file even changed, ye i can make it 3 seperate pull requests, Im gonna close this one and make 3 new ones, sorry for the bad commits

sbot50:
making 3 seperate pull requests now

PR #413, opened 2023-03-29
Add nix development shell

fgaz:
Another way to get the toolchain on any linux and osx system. I've been using this since I started using wasp-os, and it's been working fine


daniel-thompson (changes requested):
Will the new files still work if in place in tools/nix (similar to tools/docker)?

fgaz:
Yes they will, but the commands to enter a shell will have to include some paths. I pushed the change.

daniel-thompson (approved):

PR #419, opened 2023-04-21
Use `python -m` to run pytest

### LINKED ISSUE #417: "make check" returns pytest Error 127
Hi!

I'm not familiar with pytest and can't seem to make `make check` command in the master branch.

I get instantly : 
```
PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=.:wasp/boards/simulator:wasp:wasp/apps/system \
pytest-3 -v -W ignore  wasp/boards/simulator
/bin/sh: 1: pytest-3: not found
make: *** [Makefile:126: check] Error 127
```

Can anyone give me pointers on this?

fgaz:
Not every distro calls it `pytest-3`.
By using `$(PYTHON) -m` we ensure it's coherent with the python install

Closes #417

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 export PYTHONPATH := $(CURDIR)/tools/nrfutil:$(CURDIR)/tools/intelhex:$(PYTHONPATH)
 
 PYTHON ?= python3
-PYTEST ?= pytest-3
+PYTEST ?= $(PYTHON) -m pytest
 
 all : bootloader reloader micropython
 


fgaz:
On the other hand this may introduce a dependency on the working directory https://stackoverflow.com/questions/66089649/differance-between-pytest-and-python-m-pytest-tests

daniel-thompson (approved):
I think the difference in behaviour is OK (I think we would be running pytest from a makefile that requires you to be in a specific directory anyway).

Issue #425, opened 2023-06-03
Send notification via Bluetooth


jfekete1:
Hi all,

I'm trying to send a notification to my watch vie bluetooth using a Linux command line tool like bluetoothctl or gatttool.
So far I had no success.
I used gadgetbridge and sniffed the bluetooth packets sent by the android application, this is what I saw with wireshark:
Service UUID: 6e400001b5a3f393e0a9e50e24dcca9e
write command 0x52
handle 0x0013 (Nordic UART Service: Nordic UART Tx)
value 104742287b227422   ->  GB({"t"

write command 0x52
handle 0x0013 (Nordic UART Service: Nordic UART Tx)
value 3a226e6f74696679   ->  :"notify

write command 0x52
handle 0x0013 (Nordic UART Service: Nordic UART Tx)
value 222c226964223a31   ->  ","id":1

write command 0x52
handle 0x0013 (Nordic UART Service: Nordic UART Tx)
value 3638353737393230   ->  68577920

write command 0x52
handle 0x0013 (Nordic UART Service: Nordic UART Tx)
value 362c22626f647922   ->  6,"body"

write command 0x52
handle 0x0013 (Nordic UART Service: Nordic UART Tx)
value 3a2268616c69222c   ->  :"hali",

write command 0x52
handle 0x0013 (Nordic UART Service: Nordic UART Tx)
value 2273656e64657222   ->  "sender"

write command 0x52
handle 0x0013 (Nordic UART Service: Nordic UART Tx)
value 3a22c96e222c2274   ->  :" n","t

write command 0x52
handle 0x0013 (Nordic UART Service: Nordic UART Tx)
value 656c223a222b3336   ->  el":"+36

write command 0x52
handle 0x0013 (Nordic UART Service: Nordic UART Tx)
value 3730333038393532   ->  70308952

write command 0x52
handle 0x0013 (Nordic UART Service: Nordic UART Tx)
value 33227d290a         ->  4"}) 

So basically it sent this json using 11 write commands:
GB({"t":"notify","id":1685779206,"body":"hali","sender":" n","tel":"+36703089524"})

The id is the current unix timestamp.
However I'm unable to replicate this with gatttool or bluetoothctl.
Can someone provide a python script in tools folder, that can send a notification?

GaryM0101:
I'm sure you can craft a script using the PagerApp. Have a look at the pager.py in the wasp/apps/system directory, 

Cheers, Gary

jfekete1:
[...]

PR #426, opened 2023-06-07
Added send notification capability

jfekete1:


jfekete1:
Added --send-notification capability to wasptool, to fix issue #425

tools/wasptool
```diff
@@ -368,6 +368,13 @@ def handle_upload(c, fname, tname):
 
         c.expect('>>> ')
 
+def send_notification(c, title, body):
+    msg= {'title': title, 'body': body}
+    c.sendline(f'wasp.system.notifications.__setitem__(1,  {msg})')
```
fgaz, line 373: I think this will replace the notification instead of adding one. Maybe you could generate a random id or have the user specify it.

Also you should probably use `wasp.system.notify(id, {msg})`

tools/wasptool
```diff
@@ -368,6 +368,13 @@ def handle_upload(c, fname, tname):
...
+    c.sendline(f'watch.vibrator.pulse(1)')
```
fgaz, line 374:
```suggestion
    c.sendline(f'watch.vibrator.pulse(ms=system.notify_duration)')
```

PR #427, opened 2023-06-21
Add option to disable oversized button hit area

kj7rrv:
The oversized hit area can be problematic when there are many buttons close together, such as in the keyboard I'm developing. This adds an option to disable it.

### DIFF
--- a/wasp.toml
+++ b/wasp.toml
@@ -3,7 +3,7 @@
 
 [[app]]
 file = 'apps/stopwatch.py'
-quick_ring = true
+auto_load = true
 
 [[app]]
 file = 'apps/heart.py'
@@ -19,6 +19,7 @@ auto_load = true
 
 [[app]]
 file = 'apps/calculator.py'
+quick_ring = true
 
 [[app]]
 file = 'apps/disa_b_l_e.py'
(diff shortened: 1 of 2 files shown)


daniel-thompson:
I don't object but I'd don't think I want to apply this without seeing the keyboard code too.

To explain, having a button per key will consume a lot of memory (not in absolute terms but certainly relative to the 64K the watch has). It would be more memory efficient to making the keyboard grid a widget in its own right (with a single draw method and row/offset calculations to figure out which key is pressed).

Note that the calculator app did not got to the trouble of making a widget but other than that it works similar to the above. There is a draw method that draws all the keys and a touch method to figure out which key was press.

daniel-thompson:
> So you're suggesting that I draw one item on the screen, then check the tap coordinates on a grid?

Pretty much, yes.
> That might be a better option than twenty buttons. I would probably render the keyboard as an image and just blit the image to the screen?

An image is a possibility, especially if you want to use labels that are smaller than the current minimum font size. However it would be tricky to apply the colour theme to an image. The calculator is a 5x4 grid and everything is drawn from scratch (including shading to seperate the numbers and operators). It might be a good place to copy from:
https://github.com/wasp-os/wasp-os/blob/master/apps/calculator.py#L89

kj7rrv:
Thank you! I am rewriting my keyboard based on the calculator app.
KboardApp

daniel-thompson:
I've got not objections if someone wants to resurect this but I'm relunctant to merge code with no users (also some of the automatic checks were reporting problems).

PR #428, opened 2023-06-22
Add option to make wrap() ignore spaces

kj7rrv:
This adds an option to make the `wrap()` command ignore spaces.

```python
wrap("Lorem ipsum dolor sit amet", 196) # Same as old behavior
wrap("Lorem ipsum dolor sit amet", 196, True) # Same as old behavior
wrap("Lorem ipsum dolor sit amet", 196, prefer_spaces=True) # Same as old behavior
wrap("Lorem ipsum dolor sit amet", 196, False) # Still breaks on newlines, but handles spaces as regular characters
wrap("Lorem ipsum dolor sit amet", 196, prefer_spaces=False) # Still breaks on newlines, but handles spaces as regular characters
```


kj7rrv:
Thank you @GaryM0101! Is there a way to precompile the module to reduce RAM usage, or is that already done as part of compiling the micropython.zip file?

kj7rrv:
Okay, thank you @GaryM0101. My PineTime should arrive in the next couple weeks; I'll do more work on this then.
73

wasp/keyboard.py
```diff
@@ -0,0 +1,272 @@
...
+            else:
+                pass
+        else:
+            cat, value = display[self._screen][0][x + 5 * y].split(":", 1)
+
+            if cat == "key":
+                if (
+                    self._cap_mode
+                    or self._last.lower() == value
+                    and self._last[0] in "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
```
Eloitor, line 157:
```suggestion
                    and 'A' <= self._last[0] <= 'Z'
```

wasp/keyboard.py
```diff
@@ -0,0 +1,272 @@
...
+        draw = wasp.watch.drawable
+        theme = wasp.system.theme
+
+        hi = theme("bright")
+        lo = theme("mid")
+        mid = draw.lighten(lo, 2)
+        bg = draw.darken(theme("ui"), theme("contrast"))
+        bg2 = draw.darken(bg, 2)
+
+        self._draw_grid(draw, 0, bg2, lo)
```
Eloitor, line 243: If this line is moved to the end of the `open` function, just before the `self._draw()` call, then redraws for the shift or symbols keys will be somewhat faster.

thiswillbeyourgithub:
Hi. I just found out about an app called [messagEase](https://www.exideas.com/ME/index.php) whose layout was engineered for fat fingered people and smartwatches so I though it was worth sharing here for inspiration.

PR #429, opened 2023-06-23
wasp: navigate: sleep instead of vibe on swipe down

fuzzy7k:
I figure sleep is more useful than vibe. Plus, it helps minimize wear on the physical button.

### DIFF
--- a/wasp/wasp.py
+++ b/wasp/wasp.py
@@ -315,7 +315,7 @@ class Manager():
                 else:
                     # Nothing to notify... we must handle that here
                     # otherwise the display will flicker.
-                    watch.vibrator.pulse()
+                    self.sleep()
 
         elif direction == EventType.HOME or direction == EventType.BACK:
             if self.app != app_list[0]:


daniel-thompson (commented):
The user just made the gesture to show the notifications. If there are notifications then these are shown the user, if there are not any notifications pending then we pulse the motor to let the user know there are no notifications pending.

If the graphics on the PineTime was more performant when we'd probably use a visual effect to acknowledge the user's request rather than use the motor. However it's difficult to do a nice looking "visual bell" effect so we use the motor instead.

Sleeping here doesn't make sense to me. If would leave the user unable to distinguish between no notifications and the watch failing to recoginise the swiping gesture.

fuzzy7k:
Failing to recognise the swipe gesture would leave the display on. If we instantly go to sleep, that is a very good visual effect.

Plus, if there are no notifications, then what else is there to do?

Have you tried this? For me it just feels right. 
Not only does it require less effort to sleep the watch, but it reduces null brain cycles. That time between recognizing that there are no notifications and deciding what to do next. In most cases, it's sleep the watch. Already done, I don't even need to think about turning off the watch. Otherwise, I need to make the decision and bring my thumb around to push a button. A button that has an EOL. 

A bunch of micro waste is cut out with this, including battery power that would otherwise be used to pulse a motor, then continue to display a time that I've already looked at. If the user really has something else that they desire to do, the button is not far away. 

For me it has lead to more effortless use.

daniel-thompson:
Ok. So I can see what you are trying to do here. I did misunderstand it before.

However, whilst I can see I can see why it is a good personal shortcut for your workflow, I'm afraid I still really dislike it as a UI concept. It is not a particular obvious UI feature and is actually pretty unpleasent for new users will struggle to figure out why the watch turns off sometimes when they swipe down (especially if running a quick ring app that isn't the watch face).

I always just stop using the watch when I'm done. I never considered button wear, just that no decison is needed since the power off timer is very short. The battery can run the watch for many hours with the backlight on so the fifteen second timeout has very little impact on battery life. Even "acidents" with apps that don't power off aren't that problematic IMHO.

fuzzy7k:
No, it is not obvious, but neither is swiping up to bring up the launcher. It is something that people figure out pretty quick. Making it intuitive I guess, rather than obvious.

This is the first smart watch I have ever used. I didn't bother with the OS that came on it. And as a recent new user I can honestly say that I was confused why my watch was vibrating when I swiped down. Had it powered off I would have understood that as a useful feature. A vibrate served no use to me and I actually found it annoying. 

I got the idea for this from a feature I had on an old android, where double tapping the status bar would put it to sleep. I guess letting those time out is not really a good option, and wearing out the power button was, maybe still is a leading failure cause. I know that it is something that I have had to deal with on  a couple of occasions.

Anyway, I was just trying to help a bother out. If you think that new users will struggle to figure this out, we obviously disagree. But I respect that you are trying to make the right decision in your position.

daniel-thompson:
> If you think that new users will struggle to figure this out, we obviously disagree.
>But I respect that you are trying to make the right decision in your position.

Thanks for that! I don't like saying "no" but it's kinda what being maintainer means from time to time. Being open source (and using git) hopefully makes it easy to you to maintain little tweaks like this that can make the your watch entirely yours!

PR #432, opened 2023-06-29
wasp: Allow for clearing a tick request

fuzzy7k:
There currently appears to be no way to remove a tick request, and calling this func without args gives:

    self.tick_expiry = watch.rtc.get_uptime_ms() + period_ms
                       ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~
TypeError: unsupported operand type(s) for +: 'int' and 'NoneType'

### DIFF
--- a/wasp/wasp.py
+++ b/wasp/wasp.py
@@ -374,8 +374,12 @@ class Manager():
         Note: With the current simplistic timer implementation sub-second
         tick intervals are not possible.
         """
-        self.tick_period_ms = period_ms
-        self.tick_expiry = watch.rtc.get_uptime_ms() + period_ms
+        if period_ms:
+            self.tick_period_ms = period_ms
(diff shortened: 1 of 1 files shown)


fgaz:
What is the use case? An unused tick doesn't really hurt, and is often useful for keeping the top bar updated (see the timer app)

fgaz:
On the other hand it's not like this takes a ton of space. I'll defer to @daniel-thompson

fuzzy7k:
I am building a watchface with a high resolution tick and found that I get a much cleaner wakeup if I update the tick counter in the wake routine than if I wait for the next tick to update the display. This allows me to turn off ticks before going to sleep so that I do not have to do magic to handle the n-number of ticks after the tick counter has been reset.

Plus, as the commit message implies, this is also a bugfix, as calling the function with the default value of period_ms=None produces an exception at runtime.

daniel-thompson (approved):

PR #434, opened 2023-07-07
fix my own mistake: cancel alarms given an action

thiswillbeyourgithub:
A while ago I pushed a PR that was apparently not using the correct code but some old broken implementation. It was in #370

My sincerest apologies.


fgaz:
[Your signoff is missing](https://github.com/wasp-os/wasp-os/pull/434/checks?check_run_id=14866698161)

thiswillbeyourgithub:
> [Your signoff is missing](https://github.com/wasp-os/wasp-os/pull/434/checks?check_run_id=14866698161)

Done!

daniel-thompson (approved):

PR #439, opened 2023-07-15
Update appguide.rst

tomchiverton:
Clarify where to add apps to auto launch in sim.

Add notes on auto launching the sim on changes.

### DIFF
--- a/docs/appguide.rst
+++ b/docs/appguide.rst
@@ -280,9 +280,12 @@ by swiping or using the Arrow keys to bring up the launcher and then clicking
 on your application.
 
 The application can also be registered automatically when you load the
-simulator if you add it to ``wasp/main.py``. Try adding lines 5 and 6 from
+simulator if you add it to ``wasp/boards/simulator/main.pywasp/main.py``. Try adding lines 5 and 6 from
 the above example into this file (between ``import wasp`` and
 ``wasp.system.run()``).
+
+You can use something like ``while inotifywait -e modify hello.py ; do  echo change;make sim; done`` to 
+launch the simulator when you change a source file.
 
 The simulator accepts gestures such as up/down and left/right swipes but the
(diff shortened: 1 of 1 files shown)


docs/appguide.rst
```diff
@@ -280,10 +280,13 @@ by swiping or using the Arrow keys to bring up the launcher and then clicking
 on your application.
 
 The application can also be registered automatically when you load the
-simulator if you add it to ``wasp/main.py``. Try adding lines 5 and 6 from
+simulator if you add it to ``wasp/boards/simulator/main.pywasp/main.py``. Try adding lines 5 and 6 from
```
daniel-thompson, line 283: This change doens't look right.

docs/appguide.rst
```diff
@@ -280,10 +280,13 @@ by swiping or using the Arrow keys to bring up the launcher and then clicking
...
 the above example into this file (between ``import wasp`` and
 ``wasp.system.run()``).
 
+You can use something like ``while inotifywait -e modify hello.py; do   killall python3;   nohup make sim 2>&1 & done`` to
```
daniel-thompson, line 287: Nice concept. However `killall python3` is pretty brutal so I think it needs to be "Kill any existing simulator processes (and any other python3 interpretters that are currently running) and (re)launch..."

daniel-thompson (commented):
Please also squash into one patch and add a DCO signoff (see https://wasp-os.readthedocs.io/en/latest/contributing.html ).

tomchiverton:
What's a good way to spot the simulator process more precisely? Pkill isn't much help?
-- 
†øღ

fgaz:
> What's a good way to spot the simulator process more precisely? Pkill isn't much help?

It's better to used dedicated tools such as `watchexec` or `entr`. Both have options to terminate the process if it's still running.

PR #440, opened 2023-07-15
Update draw565.py

tomchiverton:
Correct order of arguments in docs

### DIFF
--- a/wasp/draw565.py
+++ b/wasp/draw565.py
@@ -459,11 +459,11 @@ class Draw565(object):
             draw = wasp.watch.drawable
             draw.line(360 / 12, 16, 64)
 
+        :param x: X coordinate of the origin
+        :param y: Y coordinate of the origin
         :param theta: Angle, in degrees
         :param r0: Radius of the start of the line
         :param y0: Radius of the end of the line
-        :param x: X coordinate of the origin
-        :param y: Y coordinate of the origin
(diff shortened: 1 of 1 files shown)


fgaz:
You'll have to sign off your commit, see [the failing check](https://github.com/wasp-os/wasp-os/pull/440/checks?check_run_id=15068556442)

tomchiverton:
Done, but this seems a high barrier to changes because edits can not be made in the Github web interface

fgaz:
Well, the github interface is only good for quick fixes anyway, and it _is_ possible to sign-off from there if you remember to. The sign-off requirement is also mentioned in the contribution docs. Finally, most changes require testing first. So in my opinion it doesn't make much of a difference

fgaz (approved):
Looks good!

I took the liberty of removing the automated "Update draw565.py" from the commit message

fgaz:
Actually it looks like you dodn't include your e-mail address in the signoff line

tomchiverton:
How do I add this from the gitlab web GUI?
-- 
†øღ

fgaz:
I have no idea. But I think I figured it out. Author and committer metadata did not match due to the first edit being made in the github web interface (one more reason to avoid it). I updated the address to the one matching the author data. It's a fake github email address, but by looking at past commits it looks like that's ok

PR #441, opened 2023-07-19
Add 15 Puzzle app

Eloitor:
[15 Puzzle](https://en.wikipedia.org/wiki/15_puzzle) app.
It also works with square boardsizes other than 4x4 such as 3x3 or 5x5, but the proportions of the blocks and font size should be adjusted.


fgaz:
Tests are failing, looks like you didn't add the app to the docs

Eloitor:
Thanks for the review, done.

daniel-thompson (approved):

PR #446, opened 2023-09-05
Adding a Four In A Row app

drtonyr:
This is the classic Four In A Row game, also known as Connect4.  You play against the computer and you can adjust the difficulty level.

This is essentially the same code as last year (2022).

Signed-off-by: Tony Robinson <tony@tonyRobinson.com>


fgaz (changes requested):
The app and the screenshot should be moved in the appropriate subdirectories

drtonyr:
> The app and the screenshot should be moved in the appropriate subdirectories

Indeed they should!

drtonyr:
> > The app and the screenshot should be moved in the appropriate subdirectories

Thanks @fgaz, the two files should be moved now.  The pushed branch on github is running on my watch, as is one where wasp.toml has my app.    The Software App calls it 'FourInARow' not 'Four In A Row', I think that's possible to fix, I just can't remember where right now.

drtonyr:
Getting the app updated to à la carte and 'make check' passing was the easy bit, getting the the Signed-off-by: is hard!

Eloitor:
Maybe you can squash the commits to a single one.

drtonyr:
Thanks @Eloitor, that was very helpful.   I learned a lot, but I'm not sure I learned enough to do this properly.   I'm still seeing one conflict, that is in README.rst.   This must be problematic, because docs/apps.rst has a similar change to README.rst, yet that doesn't show a conflict.   So I'm closer, but not there yet.

Eloitor:
Happy it worked :)

To solve the conflict, from your branch run:

1. Add the original repo as a remote. `git remote add upstream https://github.com/wasp-os/wasp-os` (You can name it as you like)
2. Get the last changes: `git fetch upstream master`
3. `git rebase upstream/master`

If you get conflicts solve them.

drtonyr:
Thanks @fgaz and @Eloitor for all your help.  I learned a lot, it just turned out easier to start again.

PR #452, opened 2023-09-18
fix: simulator was missing some time functions

thiswillbeyourgithub:
Signed-off-by: thiswillbeyourgithub <26625900+thiswillbeyourgithub@users.noreply.github.com>

### DIFF
--- a/wasp/boards/simulator/watch.py
+++ b/wasp/boards/simulator/watch.py
@@ -5,6 +5,9 @@ import time
 def sleep_ms(ms):
     time.sleep(ms / 1000)
 time.sleep_ms = sleep_ms
+time.ticks_ms = lambda : int(time.time() * 1000)
+time.ticks_us = lambda : int(time.time() * 1000 * 1000)
+time.ticks_diff = lambda x, y : x-y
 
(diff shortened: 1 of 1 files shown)


wasp/boards/simulator/watch.py
```diff
@@ -5,6 +5,9 @@
 def sleep_ms(ms):
     time.sleep(ms / 1000)
 time.sleep_ms = sleep_ms
+time.ticks_ms = lambda : int(time.time() * 1000)
+time.ticks_us = lambda : int(time.time() * 1000 * 1000)
+time.ticks_diff = lambda x, y : x-y
```
fgaz, line 10: Looks like these are a bit more complex than simple subtraction and multiplication: https://docs.micropython.org/en/latest/library/time.html#time.ticks_diff https://docs.micropython.org/en/latest/library/time.html#time.ticks_ms

fgaz: Nevermind, that's just wraparound, which shouldn't be a problem with python's arbitrary-precision integers

fgaz (approved):

Issue #445, opened 2023-09-02
more deepsleep ?


thiswillbeyourgithub:
Hi,

Not very knowledgeable about any of this so I preferred to ask. Recently Infinitie managed to get more than 10 day of battery life which is dramatically longer than what we can currently get out of the waspos.

My question is : is there a way to modify the balance between usability and battery? I would be very interested in a less responsive watch that spends more time in deep sleep for example.

Additionnaly, is anyone interested by listing the different milestones needed to improve battery life? I think I remember something about the bluetooth beacon being always on for example?

Cheers

GaryM0101:
Infinitime released a battery fix in their latest release. Just tested
it on the Pinetime and got 28 days till the battery got to 18%.
Impressive! 

On 2023-09-02 11:43, thiswillbeyourgithub wrote:

> Hi, 
> 
> Not very knowledgeable about any of this so I preferred to ask. Recently Infinitie managed to get more than 10 day of battery life which is dramatically longer than what we can currently get out of the waspos. 
> 
> My question is : is there a way to modify the balance between usability and battery? I would be very interested in a less responsive watch that spends more time in deep sleep for example. 
> 
> Additionnaly, is anyone interested by listing the different milestones needed to improve battery life? I think I remember something about the bluetooth beacon being always on for example? 
> 
> Cheers 
> 
> --
> Reply to this email directly, view it on GitHub [1], or unsubscribe [2].
> You are receiving this because you are subscribed to this thread.Message ID: ***@***.***>
 

Links:
------
[1] https://github.com/wasp-os/wasp-os/issues/445
[2]

thiswillbeyourgithub:
I addde a draft pr at #453 for anyone interested in trying to make the watch less responsive but lasting longer.

PR #454, opened 2023-09-18
WIP: improving battery life by sleeping more

thiswillbeyourgithub:
Hi, (sorry for botching my previous PR at #453)

This is just a draft related to #445 where I show what I'm currently testing to try to save a bit of battery life.

My view is that I want a less responsive watch that lasts longer.

Any help is much welcome.

Signed-off-by: thiswillbeyourgithub <26625900+thiswillbeyourgithub@users.noreply.github.com>

### DIFF
--- a/wasp/boards/simulator/watch.py
+++ b/wasp/boards/simulator/watch.py
@@ -5,6 +5,9 @@ import time
 def sleep_ms(ms):
     time.sleep(ms / 1000)
 time.sleep_ms = sleep_ms
+time.ticks_ms = lambda : int(time.time() * 1000)
+time.ticks_us = lambda : int(time.time() * 1000 * 1000)
+time.ticks_diff = lambda x, y : x-y
 
 import sys, traceback
 def print_exception(exc, file=sys.stdout):
--- a/wasp/wasp.py
+++ b/wasp/wasp.py
@@ -503,9 +503,10 @@ class Manager():
                 while time.ticks_diff(time.ticks_ms(), bef) <= _INTERVAL:
                     machine.deepsleep()
                 if self._button.get_event():
-                    return self.wake()
+                    self.wake()
+                    return
                 elif alarms and rtc.update() and rtc.time() >= alarms[0][0]:
-                    return self._tick()
+                    return
 
     def run(self, no_except=True):
         """Run the system manager synchronously.
--- a/wasp/wasp.py
+++ b/wasp/wasp.py
@@ -22,10 +22,14 @@ import sys
 import watch
 import widgets
 import appregistry
+import os
+import time
 
 from apps.system.launcher import LauncherApp
 from apps.system.pager import PagerApp, CrashApp, NotificationApp
 from apps.system.step_counter import StepCounterApp
+
+_INTERVAL = micropython.const(1000)
 
 class EventType():
     """Enumerated interface actions.
@@ -466,13 +470,8 @@ class Manager():
         update = rtc.update()
 
         alarms = self._alarms
-        if update and alarms:
-            now = rtc.time()
-            head = alarms[0]
-
-            if head[0] <= now:
-                alarms.remove(head)
-                head[1]()
+        if update and alarms and alarms[0][0] <= rtc.time():
+            alarms.pop(0)[1]()
 
         if self.sleep_at:
             if update and self.tick_expiry:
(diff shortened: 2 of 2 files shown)


beardeddude:
I tried to get this running on my pinetime this morning. I cant get it to wake up after it sleeps. If I press the button long enough it will still reset and I can reboot it.

thiswillbeyourgithub:
You have to press a bit longer. Max 1s. Try clicking few more times

beardeddude:
I tried pressing it for a long time at different intervals. Up to to the point of resetting. You also have a pinetime? I can play with it some more later

thiswillbeyourgithub:
I do have a pintime. Maybe try setting _INTERVALL to 500?

beardeddude:
After messing around for a bit I got it working. The alarm loop was the issue but im not sure why. But that logic is not needed anyways. I removed the alarm else-if and just left the else block and then removed the infinite loop from that block so it looked like below. Its equivalent because _tick is called in a loop. So its ok to just loop to interval time and just let the method continue.

`else:  # sleeping without alarms: sleep until button pressed
            bef = time.ticks_ms()
            while time.ticks_diff(time.ticks_ms(), bef) <= _INTERVAL:
                machine.deepsleep()
            if self._button.get_event():
                self.wake()
                return`

I'll test this on my daily driver and see if it results in a longer battery life. I think if we want to take this any further it is going to require changes at a lower level than the python code. There is a note in the code from Thompson stating that the power save logic needs work and that it will cause issues in the sim (_run method) if fixed. Still its worth a shot to just call deepsleep and see what happens. Otherwise we would probably need to look at things like changing the cpu frequency when sleeping, setting the interrupts for the deepsleep command and getting deepsleep to stay in effect longer, and bluetooth tweeks, etc.

thiswillbeyourgithub:
Thanks I'lm take a look later. I noticed that I actually have an issue when the alarm is on I can't get it out of sleep either. Will try your fix.

IIRC the main power draw is that the bluetooth is always on.

While coding I also noticed that calling deepsleep with an argument throws an exception whereas normally it takes an optionnal argument for how long to sleep. This would avoid the need for my workaround.

Lastly I couldn't get machine.Timer's callback to work. If working on the watch that too would make the _tick logic lighter IMO.

But I know 0 low level code so I can't do much.

thiswillbeyourgithub:
I thought some more and I think your fix cannot work. Your example will keep sleeping until a button is pressed, even if an alarm is overdue. I did a compromise in a single loop but I'm not sure it's better because it checks every _INTERVAL if there are any alarms, even though new alarms cannot be activated while the screen is off, hence why I did 2 separate loops in the beginning.

But that still does not explain why the watch seemingly does not wake up when there is an alarm (at least in the code we tried, don't know yet about my most recent push)

What do you think?

thiswillbeyourgithub:
Alright. That's not within my competence unfortunately so my quest stops here. Thanks!

daniel-thompson:
> Alright. That's not within my competence unfortunately so my quest stops here. Thanks!

I guess I'll close the PR at this point then.

FWIW I did do some work to study the wake ups per second (which is the metric you care about for this type of work) a *long* time ago. Even when python it doing nothing the watch is still wakes up hundreds of times per second (e.g. >99.9% of all wakeups) and I assume this is due to the SoftDevice bluetooth stack meaning it can't be changed from the python code.

The most interesting place to start looking is definitely the C code. In particular the C code that implemnets deepsleep() looks like it is calling the WFI (wait-for-interrupt) instruction instead of telling the SoftDevice to enter a low power state. If anyone does want to try tuning power this is one area to look at (the other is to switch micropyhon over to a different BLE stack, which is an even bigger job).

thiswillbeyourgithub:
Thanks. One question comes to mind : if I only use the bluetooth to flash wasp os and never directly when wasp is booted, is it possible to disable the bt stack in a way that wouldn't brick the watch but save battery?

beardeddude:
The disaBLE app will do that

thiswillbeyourgithub:
Oh really ? I thought disaBLE would not save battery

GaryM0101:
True in my case, in fact using disaBLE uses more battery than having BLE
enabled im my two watches.

On 2023-10-07 09:16, thiswillbeyourgithub wrote:

> Oh really ? I thought disaBLE would not save battery 
> 
> --
> Reply to this email directly, view it on GitHub [1], or unsubscribe [2].
> You are receiving this because you are subscribed to this thread.Message ID: ***@***.***>
 

Links:
------
[1] https://github.com/wasp-os/wasp-os/pull/454#issuecomment-1751748807
[2]

thiswillbeyourgithub:
edit just in case someone has my issue: actually what was messing up my alarms was me forgetting to handle the case where alarms are set and the button is pressed : 

```
        elif not alarms:  # screen off and no alarms: sleep profoundly
            while True:
                bef = time.ticks_ms()
                while time.ticks_diff(time.ticks_ms(), bef) <= _INTERVAL:
                    machine.deepsleep()
                if self._button.get_event():
                    self.wake()
                    break
        else:  # I forgot that :/
            if self._button.get_event():
                self.wake()

```

PR #457, opened 2023-09-22
Minor change to wrap() to allow character wrapping

drtonyr:
This is a minor change to wrap() to take an optional argument.

If called with char=True then the code will wrap at character boundaries, not word boundaries.  This is useful in many places, especially where truncation is necessary as it reduces the amount of lost information.


drtonyr:
I'm going to close this, I don't fully understand wrap() as it stands.  I'll come back when I do, there's plenty of other work to be done...

PR #459, opened 2023-09-23
Small change to wasp.watch.drawable.string() so that text can be left justified.

drtonyr:
When width is set we can right and centre justify text, but the functionality for left justification was missing.

Example, running in the simulator:
```
>>> draw.fill(0xffff, 0, 0, 240, 240)
>>> draw.string('centre', 40, 40, width=160)
>>> draw.string('right', 40, 100, width=160, right=True)
>>> draw.string('left', 40, 160, width=160, right=False)
```
Screenshot_2023-09-23_09-33-41

It's worth noting the uses of the 'right' parameter at the moment:
```
apps/calculator.py:        wasp.watch.drawable.string(output, 0, 14, width=200, right=True)
apps/pomodoro.py:            draw.string(self.queue, 0, 35, right=True, width=240)
apps/pomodoro.py:            draw.string(self.queue, 0, 35, right=True, width=240)
apps/pomodoro.py:            draw.string(self.queue, 0, 35, right=True, width=240)
```

There were some implementation choices that are worth discussing.  The desired functionality is ternary, there are three choices, 'left', 'centre' and 'right'.  If starting from scratch, a clean solution would to have a justify= parameter which took  a string variable.   However, we are not starting from scratch, there is backwards compatibility to be maintained, and also string variables take space, and we don't have much space.   So this leads to keeping the 'right' parameter and making it ternary, 'None' is the obvious value to add.  Now there are two choices:

1. right=None is left justified, right=False is centre justified.  This is 100% backward compatible.
2. right=False is left justified, right=None is centre justified.  It is possible, but very unlikely, that someone has code that uses right=False already, and this patch would break that code.

The second option is much cleaner to me.  The default is centre justification, by specifying 'right' we are saying how we want to override this default.  False is the opposite of True, so the opposite of right justification is left justification.  Calling  wasp.watch.drawable.string() with right=None to get left justification is quite counter intuitive to me.


daniel-thompson:
I'd be tempted to use -1/0/1 for the "ternary" (it conveys left/centre/right more intuitively than None/False/True).

You can even call the argument "justify" since we can either:

1. Fix the built-in apps and assume no important out-of-tree apps are using right=True
2. Deprecate right and change the default value of right to None, then add:
~~~python
if right is not None:
    justify = int(right)
~~~

daniel-thompson:
PS Use comparisons to zero to the logic isn't ambiguous: justify < 0, justify == 0, justify > 0 .

drtonyr:
Yes, sure, we both agree that justify=int is cleaner.

I don't know how you deprecate an option.  If you accept this, I'd be happy to make another pull request that changes all right= to justify=  (which will change the docs) in master.   Accepting that will get us to the cleaner solution pretty soon.

drtonyr:
Now with justify=int and the other code suggested above.  param right marked as deprecated.  I'd be happy to change all right=True to justify=1 in a further PR if this is accepted.

wasp/draw565.py
```diff
@@ -300,20 +300,23 @@ def string(self, s, x, y, width=None, right=None):
                       fill the full width with the background colour (to ensure
                       that if we update one string with a narrower one there is no
                       need to "undraw" it).
-        :param right: Set the justification mode.  The default is to centre, if
-                      set to True then right justify, if False then left justify.
+        :param right: Deprecated: use justify=1 instead
```
daniel-thompson, line 303: We don't have to nuke the documentation on how right works. Better to keep the original docs and add that is deprecated and replaced by justify.

daniel-thompson (commented):
Just one comment now.

Note that it might be easier to `git rebase` and squash/fixup the history rather than pile new commits on top...

drtonyr:
Agreed, thanks.   I've reverted the documentation changes to be the least possible.

Unfortunately it seems I've made a mess of this branch.  I've spent a few hours on it and it's only getting worse, I don't think I have an option but to do it again cleanly.

Issue #251, opened 2021-09-11
wrap() does not always work with string(width=...)


fgaz:
works:

```python
s = "testwithareallylongname"
draw.string(s[:(draw.wrap(s, 240)[1])], 0, 200)
```

crashes:

```python
s = "testwithareallylongname"
draw.string(s[:(draw.wrap(s, 240)[1])], 0, 200, width=240)
```

trace:

```
[omitted]
    draw.string(s[:(draw.wrap(s, 240)[1])], 0, 200, width=240)
  File "/home/fgaz/src/wasp-os/wasp/draw565.py", line 321, in string
    self.fill(bg, x, y, leftpad, h)
  File "/home/fgaz/src/wasp-os/wasp/draw565.py", line 153, in fill
    display.set_window(x, y, w, h)
  File "/home/fgaz/src/wasp-os/wasp/drivers/st7789.py", line 132, in set_window
    window[2] = xp >> 8
ValueError: byte must be in range(0, 256)

**Comment by drtonyr** [comment]
I was just about to open an issue with my own debug example then I found this one (from 2 years ago).

I just started on improving wrap(), then gave up because I was getting similar errors.   The problem isn't with wrap, it's much lower down.  The problem is that some calls to draw.fill() that have zero pixels fail because display.set_window() can't cope with zero pixels when x = 0.

st7789.py contains the code:
```
  xp = x + width - 1
  ...
  window[2] = xp >> 8
```
We've both managed to trigger a strange edge case.  When both x is zero and the window width is zero then xp is -1 and `-1 >> 8` isn't in the range (0,256).  To see this here is the guts of the code that fails:
```
>>> draw.string('testwithareallylong', 0, 200, width=240)
```
but just indent by one pixel and it works:
```
>>> draw.string('testwithareallylong', 1, 200, width=239)
```

Now, the question is where to fix it?   There seems to be two choices:

1. fill(bg, x=0, y=0, w=None, h=None) - simply return if w*h is zero, before set_window() is called, this works
2. set_window(x, y, width, height) - this is low level, I don't understand this code.  Using `xp = max(x + width - 1, 0` seems to work.

both apps/gallery.py and apps/game_of_life.py call set_window().   Is it reasonable to call set_window() with zero width or height and expect it to work?   I don't know.

[...]

PR #460, opened 2023-09-24
Small change to fill() to exit early if there is nothing to do.

drtonyr:
This avoids running unnecessary code and avoids a bug in set_window().

For background see #251

### DIFF
--- a/wasp/draw565.py
+++ b/wasp/draw565.py
@@ -150,9 +150,11 @@ class Draw565(object):
         if h is None:
             h = display.height - y
 
-        display.set_window(x, y, w, h)
-
         remaining = w * h
+        if remaining == 0:
+          return
+
+        display.set_window(x, y, w, h)
 
         # Populate the line buffer
         buf = display.linebuffer


daniel-thompson:
This change is, IMHO, somewhat different to the .set_window() case because, arguably, there is a correct way to interpret how to draw a line that is zero pixels wide (or high). However I'm still on the fence about it because drawing a zero pixel wide line still isn't useful (e.g. it's probably still a bug in the caller).

Does this change result in any changes to the selt-test drawing performance benchmarks, especially the text rendering ones?

drtonyr:
timings for make check:

master: 245 passed in 66.26s, 65.86s, 66.62s, 67.77s  -- average 66.6275 +- 0.712298
fill: 245 passed in 66.72s, 67.48s, 68.08s, 66.65s  -- average 67.2325 +- 0.587681

That's inconclusive.  How do I run just the  text rendering benchmarks?

We need to catch an edge case that the hardware can't cope with.  This isn't done at the moment, any fix will be more code an slightly slower.  This fix uses an existing variable and exits the fill() call early if it's zero, you can't get more lightweight than that.   There won't be many edge cases in the existing code, and none in the tests as then the tests will fail, so the tests won't show any potential speed up, this patch will just try to catch cases that aren't in the tests.

It's not really a bug in the caller.  Where we have a problem is cases like when we want to draw text in a window, we don't know if we have a zero width pad or not.   We could check for this in many places in the code or we could just say that if we are called and the answer happens to be zero then do nothing.  Doing it this way makes for cleaner code, we don't have to check for zero elsewhere, less code elsewhere, less to go wrong, more Pythonic this way.

daniel-thompson:
To get benchmarks, enable the Self Test application, fire it up and scroll up/down through the tests. Performance in the line drawing (both vertical and angled) and text rendering tests are going to be the most interesting. Note that they to run on a real watch, not the simulator)

In this case, I'm reasonably happy from an API point of view to allow zero sized fills to do nothing, but as mentioned in the other ticket, drawing on wasp-os is very heavily optimized and this is a part of the code base where we might be willing to trade 100% pythonic for performance (in short I'm happy to take this if the costs on the benchmarks are not too bad... and in this case I don't know whether or not they will be).

drtonyr:
Thanks.  Here are the number from my P8 (there is only one line test):

Orig Fill-H: all start 0.081 then 62 81 82 62 81 86 69 75  so 0.08175 +- 0.00008
Orig Fill-V: all start 0.081 then 85 67 88 67 86 86 59 65 so 0.08175 +- 0.00011
Orig line: all start 0.340 then 84, 78, 99, 79, 76, 86, 78, 71 so 0.34081 +- 0.00008
Orig string: all start 0.3 then 341, 297, 299, 299, 297, 298, 298, 297 so 0.3303 +- 0.0014

PR Fill-H: all start 0.08 then 195, 197, 204, 199, 204, 192, 195, 218 so 0.08200 +- 0.00007
PR Fill-V: all start 0.08 then 230 247 225 244 229 237 251 251 so 0.08239 +- 0.00010
PR line: all start 0.34 then 277 256 262 28 274 257 282 254 so 0.34268 +- 0.00010
PR string: all start 0.3 then 340 300 297 298 297 298 300 297 so 0.3292 +- 0.0035

It's statistically significant in Fill-H and Fill-V but it's also less than a millisecond and less than 1% slower, you've got to do a lot of work to notice this difference.  line is also less than a percent slower, string is the same speed.

Note that these tests demonstrate the bug I'd like to fix, that is if the time is drawn with a lot of decimal places on one test and subsequently with fewer then the end of the last results are visible.   It would be great if we could use width= more.

If you don't want to fix it here then I'll do a PR where I fix up string, specifically if leftpad/rightpad is zero then we won't call fill and we won't throw a wobbly on a valid call.  I can then continue and if we find other places that are problematic then we can fix them later.

Finally, note that we can call fill() and it not be obvious that we are making a zero height or width request, if x or y is 240 and w/h not specified then they are auto calculated to zero and we have a problem.

daniel-thompson (approved):

PR #461, opened 2023-09-24
set_window(): Avoid an exception when x is zero and the width is zero

drtonyr:
We have a bug in set_window() that is triggered when x is zero and the width is zero or y is zero and the height is zero.

You could say "well don't do that then", but this situation does happen when we asked for strings to be displayed in a fixed width, see #251 for background.  There is a work around for that particular case in #460.   However, this patch makes the system more robust to direct calls to set_window().


daniel-thompson:
> You could say "well don't do that then"

I've been thinking about this and I am inclined to say exactly that!

To explain, throwing an exception when a function is asked to do something nonsensical (such as set a hardware window to value that cannot be represented by the hardware) is "not a bug" because it is arguably the correct thing to do. In this case it is impossible to figure our what the caller wanted to happen. Python exceptions are how we communicate to the caller that the current function cannot figure out how to safely proceed and that the current function needs help from one of the parent functions to figure out what to do next.

Now, I admit we got lucky with the exception since the code wasn't intentionally designed to raise an exception here. However IMHO improving the docs to make it clearer what the preconditions of the function are is better than trying to avoid the exception here: the current function's behavior is a sensible defensive measure against abuse by callers (and is free in terms of code size ;-) ).

drtonyr:
Well I mostly agree, but I don't think we are asking for something nonsensical, just something that can't be represented by a hardware call.   It's not up to the Micropython developer to know the hardware, we are supposed to abstract the details.

However, I'm not certain whether a zero width or zero height window has a real use or not, I just don't have the experience at this level.   You can compare this to the fill() call where I do have experience, in that case you could also say that it's nonsensical to fill a rectangle of zero width or zero height and it certainly can't be represented by a hardware call, but we do it all over the code.  In the case of fill() the Pythonic thing to do is just to accept zero width or zero height and do nothing.   You could definitely make the same argument for set_window().

Anyway, I made two pull requests so that this bug would get fixed one way or another.  If you accept the other one, let's close this one.  If set_window() becomes problematic for anyone they will search for it and find this thread.  In the meantime everything we want to work will work.

daniel-thompson:
> In the case of fill() the Pythonic thing to do is just to accept zero width or zero height
> and do nothing. You could definitely make the same argument for set_window().

I don't think such an argument works for set_window() because it is unclear what to do next. If .set_window() ignores it arguemtns and exits successful with the hardware set of a window that was not requested then what is the driver supposed to do when the caller writes data?

IMHO throwing an exception here is "the right thing to do". Naturally it would be more pythonic to throw a more exact exact exception (DrawingError?) since that would be easier to catch. However, I don't think we can spare the performance for that! The drawing functions are the most heavily optimized part of the entire of wasp-os. The watch would be unusable without these optimizations but even with them, most of the latency during an application switch is spent redrawing the screen. Thus I think we should be OK with being 75% pythonic and keep the benchmarks from being impacted.

drtonyr:
Yes, you are right, I've done some digging on the ST7789 wrt CASET and RASET.   I now understand that the way it's designed is that it needs start and end positions, both of which are used.  This means that there are two cases that we call where the call is out of bounds for what the hardware can support.  In both cases the width or height is zero.

1. the start (x/y) is zero.   Here we throw an exception, which is the right thing to do (and agreed, we don't want a more detailed exception).  This happens as the computed end position is -1 which we can't represent.
2. the start (x/y) is non-zero.  Here we do not throw an exception, but we are setting the hardware into a bad state.

We get away with (2) routinely - it's the result of a fill() call for which the only write must be zero bytes and it seems that zero byte writes with dud windows are tolerated.

My patch did not ignore the invalid state causing problems later, it forced invalid calls to be valid.   However, this is still the wrong thing to do.   

As you say, we can't afford to error check at this level.   We can document that whilst we don't catch it, the hardware can not cope with width or height being zero.  Currently this happens routinely when rightpad=0 in string().  We can try harder to not set the hardware into an invalid state.

Incidentally, whilst researching this I found http://www.technoblogy.com/show?3UEJ which says that you can read from the display and therefore do smooth animations with the XOR trick.   This is just for fun, I'm not suggesting we implement in micropython, but it's interesting.

PR #465, opened 2023-10-02
Wasp freezes on the main.py screen - Installation

GaryM0101:
Most new users of WASP-OS do not know the limitations and may attempt to autoload a number of applications that exceed the capacity to freeze into the WASP-OS binary. This results in the watch freezing on the main.py screen. So the user will have to reduce the number of applications chosen and re-install.


docs/install.rst
```diff
@@ -587,6 +587,15 @@ Most problems with normal mode operation occur either because ``main.py`` is
 missing, out-of-date or corrupt. These issues most commonly result in an
 entirely black screen when running the watch is running in normal mode.
 
+Wasp freezes on the ``main.py`` screen
```
daniel-thompson, line 590: I'd recommend merging this a little with the existing docs since I think the result would read better. For example:

~~~
Most problems with normal mode operation occur either because ``main.py`` is
missing, out-of-date or corrupt, or because too many applications are being started by default, resulting in the system running out of RAM.

Out of memory problems are best addressed by reducing the number of applications you have "frozen" into the wasp-os binary. <glue in your text from here>.
~~~

GaryM0101:
Closing this one per my commit missteps, PR 468 created in lieu of.

PR #469, opened 2023-10-08
add justify= to .string() and deprecate right=

drtonyr:
Unfortunately I got in a right mess with #459 and no longer had the confidence I was doing the right thing.

This PR addresses the known issues with #459, namely it makes minimal changes to the documentation and is a single commit.

### DIFF
--- a/wasp/draw565.py
+++ b/wasp/draw565.py
@@ -292,34 +292,42 @@ class Draw565(object):
         """
         self._font = font
 
-    def string(self, s, x, y, width=None, right=False):
+    def string(self, s, x, y, width=None, right=None, justify=0):
         """Draw a string at the supplied position.
 
         :param s:     String to render
         :param x:     X coordinate for the left-most pixels in the image
         :param y:     Y coordinate for the top-most pixels in the image
-        :param width: If no width is provided then the text will be left
-                      justified, otherwise the text will be centred within the
-                      provided width and, importantly, the remaining width will
-                      be filled with the background colour (to ensure that if
-                      we update one string with a narrower one there is no
+        :param width: If no width is provided then the text will start at the
+                      coordinates provided, otherwise the text will be justified
+                      within the provided width and, importantly, the remaining
+                      width will be filled with the background colour (to ensure
+                      that if we update one string with a narrower one there is no
                       need to "undraw" it)
-        :param right: If True (and width is set) then right justify rather than
-                      centre the text
+        :param right: Deprecated: If True (and width is set) then right justify
+                      rather than centre the text
+        :param justify: Set the justification mode.  The default (0) is to centre,
+                      if set to 1 then right justify, if -1 then left justify.
         """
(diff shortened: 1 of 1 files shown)


drtonyr:
bump

Yes, I know I screwed up with the last PR (#459) getting everything into one merge.   However, I don't think this is a complex issue and I do think I addressed all concerns.

Let me try to make a strong case for this PR.  Screen drawing is slow, if we call fill() and start again it's visible.  If we want to update just part of the screen with text then we really don't want to blank it and write an unknown length, we really want to make one call with a width and it simply overwrites what's there.  So all calls to .string() really should have width= set and we can't do that at the moment if we want left justification.

I have a watch face which also displays the tide times for holiday use.  When I switch location it used to blank the screen and redrew at a new location which was very noticeable.  I've just updated it so that it redraws on the old screen with the new location data, it looks so much nicer.  Unfortunately I don't have  left justification - it's waiting on this PR (not that I'd be offended if nobody wanted my tide times - it is niche).

GaryM0101:
Tide times...now that would be nice! I thought Daniel passed you the
torch to merge this and other PR's into the current source, or did I
miss something?

On 2023-11-09 06:44, Tony Robinson wrote:

> bump 
> 
> Yes, I know I screwed up with the last PR (#459 [1]) getting everything into one merge. However, I don't think this is a complex issue and I do think I addressed all concerns. 
> 
> Let me try to make a strong case for this PR. Screen drawing is slow, if we call fill() and start again it's visible. If we want to update just part of the screen with text then we really don't want to blank it and write an unknown length, we really want to make one call with a width and it simply overwrites what's there. So all calls to .string() really should have width= set and we can't do that at the moment if we want left justification. 
> 
> I have a watch face which also displays the tide times for holiday use. When I switch location it used to blank the screen and redrew at a new location which was very noticeable. I've just updated it so that it redraws on the old screen with the new location data, it looks so much nicer. Unfortunately I don't have left justification - it's waiting on this PR (not that I'd be offended if nobody wanted my tide times - it is niche). 
> 
> --
> Reply to this email directly, view it on GitHub [2], or unsubscribe [3].
> You are receiving this because you are subscribed to this thread.Message ID: ***@***.***>
 

Links:
------
[1] https://github.com/wasp-os/wasp-os/pull/459
[2] https://github.com/wasp-os/wasp-os/pull/469#issuecomment-1803965070
[3]

fgaz:
> I thought Daniel passed you the torch to merge this and other PR's into the current source, or did I miss something?

Conditional on one approval, which is missing. Let me have a look...

drtonyr:
@GaryM0101 Tides has been a lot of fun.  I started in early September after returning from a holiday house that had a broken tide clock.  I thought it would only take a couple of weeks to do something much better.  Well, it is much better, just it took me a couple of months.   I'll write it up in the PR, there's a fair amount to talk about.

@fgaz Thanks!

Issue #458, opened 2023-09-22
/flash - how should we lay out the directory structure?


drtonyr:
We are often short on space, but in fact we have oodles of space in /flash (4MB SPI NOR XT25F32B), just it's not commonly used.

At the moment we have main.py and a logs directory.   I'd like to see a policy that gives us a clean filing system, which means that apps have their own space.

I have a strong preference for not using plurals in names (e.g. file names or variable names), but WASP-OS doesn't, so to keep things simple I'll only suggest plurals.

That said, here are a few ideas.  Firstly we need a naming convention for apps, here are some choices following [app-naming-conventions-and-placement](https://wasp-os.readthedocs.io/en/latest/appguide.html#app-naming-conventions-and-placement)

1. the file name in snake_case
2. the class name in pascal case
3. the class name with App appended
4. the value of the NAME variable

I'm in favour of (2), it's the one that is used in most places.  Whatever is chosen, lets call it $APP.   Now we have some choices as to how we lay out the file system:

1. top level - /flash/$APP.$EXT for one or more values of $EXT, e.g. log, cfg, etc.
2. distributed according to type, /flash/$EXT/$APP.*.   This is what we have now with /flash/logs, we could also have /flash/etc or /flash/cfg
3. distributed according to APP with no top level directory, /flash/$APP/* - developers can put what they like under  /flash/$APP/
4. distributed according to APP with a top level directory /flash/apps/$APP.$EXT or /flash/apps/$APP/*.   If only one file is needed then developers can use /flash/apps/$APP.$EXT, if more than one then they must create a directory
5. distributed according to APP with an enforced directory /flash/apps/$APP/*

I'm in favour of (5), it has the advantage that within the directory fixed names could be used, like /flash/apps/$APP/ICON.rle and the developer is pretty much free to do what they want in this space, which is important as all apps are different.  Flash space is accountable for (not the case if files are distributed) and can be easily cleaned up.  I'd also be happy with (3).

To give some concrete examples, I'm currently working on TideClock that needs a text location database.   So I could have /flash/apps/TideClock/locations.txt where users could enable (uncomment) one or a couple of favourite locations and the clock face could scroll through these.   I've also go a 'working' editor, if people want it then I'll need some default space to create new files and I'd do that with /flash/apps/EditExec/$DATE.txt.   I also like the idea of reading the icons from /flash, but I haven't done any timings.

[...]

PR #470, opened 2023-10-08
Bible app

kj7rrv:
This PR adds a Bible app.


drtonyr:
> This PR adds a Bible app.

My first thought is to ask how much space this takes up.  Have you run it on a watch and if so what else fits into the memory at the same time?

kj7rrv:
> My first thought is to ask how much space this takes up.  Have you run it on a watch and if so what else fits into the memory at the same time?

It's too early in development to really determine that. The text is stored in flash, and the app won't be able to store much in memory at one time. All this really is right now is a text retrieval library; there's no interface yet.

drtonyr:
> My first thought is to ask how much space this takes up. Have you run it on a watch and if so what else fits into the memory at the same time?
> It's too early in development to really determine that. The text is stored in flash, and the app won't be able to store much in memory at one time. All this really is right now is a text retrieval library; there's no interface yet.

Sorry for deleting my reply, I didn't read your code fully and thought I'd cover up my ignorance!

More thoughts:
- you might like to contribute to #458.  It's an open issue as to how /flash is organised, but I don't see that we'll end up with a top level dir for each app
- I see that you are looking to include an index with each file, great.  You'll only have a few kbyte memory to play with, so you'll probably want to index down to lines as you can't display more than a few lines anyway.
- you might want to remove _ALL_BOOKS and replace by a call to os.listdir (perhaps with memory freeing), that way you use less flash and you can add other books
- commits need to be signed - https://wasp-os.readthedocs.io/en/latest/contributing.html?highlight=sign

kj7rrv:
@drtonyr Thank you! I'll take a look at the flash layout discussion. I'd thought about indexing per line, but my current plan is to have the pager read out only as much as it can show per page. I had tried os.listdir(), but was consistently getting MemoryErrors when using it, so I switched to a predefined list. The only reason I can think of to add more books would be to include the Deuterocanon/Apocrypha, and if anyone is interested in that, then those books could be added to the predefined list.

Regarding signing, all I'm seeing a requirement for is signing off, which I believe I did?

drtonyr:
> @drtonyr Thank you! I'll take a look at the flash layout discussion.

Thanks, I'll follow up there.

> I'd thought about indexing per line, but my current plan is to have the pager read out only as much as it can show per page.

That sounds right.  I couldn't tell how big the chunks were from the index values at the start, I guessed wrong.

> I had tried os.listdir(), but was consistently getting MemoryErrors when using it, so I switched to a predefined list. The only reason I can think of to add more books would be to include the Deuterocanon/Apocrypha, and if anyone is interested in that, then those books could be added to the predefined list.

Fair enough.  We have to balance code storage and RAM, it's not easy!   Your directory structure had KJV in it so I guessed that you might be interested in adding other versions.

> Regarding signing, all I'm seeing a requirement for is signing off, which I believe I did?

Indeed you did - I see 'Signed-off-by: Samuel Sloniker <sam@kj7rrv.com>' which looks just right.  However, at the top of this page I also see 'unverified' and when I click on it then it says 'This commit is not signed, but one or more authors requires that any commit attributed to them is signed.' .  Looks like I got confused by the message, there's definitely another step to find here as I don't get that.

PR #478, opened 2023-10-30
Fix docs issues

### LINKED ISSUE #477: docs: appguide: Application life-cycle -- fix lifecycle diagram
The application life-cycle diagram (png file) got corrupted. I sourced a copy from the wayback archive to repopulate.

fgaz:
Looks like recommonmark is needed https://readthedocs.org/projects/wasp-os/builds/22389134/

Also fixes #477

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -7,6 +7,9 @@ build:
   os: ubuntu-22.04
   tools:
     python: "3"
+  apt_packages:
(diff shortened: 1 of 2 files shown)


fgaz:
Maybe we should freeze the versions in requirements.txt to avoid future breakage, but let's merge this for now

daniel-thompson (approved):

PR #494, opened 2024-04-20
Fix build up to GCC 13

fgaz:


fgaz:
The -Wno-* options do not exist on gcc 11, I should have known. Maybe I should just use `-Wno-error` since we're not working on micropython itself anyway.

daniel-thompson (approved):

PR #502, opened 2024-08-05
Create Guidelines for UI and Migrate Apps and Faces to self contained folders

BeatLink:
This PR serves two purposes:

1. To create an initial design guideline for user interfaces on the OS and 
2.  To migrate apps and watch faces to self contained folders for easier packaging and development. 

As part of this, the tools/configure_wasp_apps.py was heavily refactored to match the new file format as well as to simplify the code especially relating to string concatenation and file content structuring. 

The new packaging format is as follows

* Every app or face must be stored in a folder
* The name of the folder is used as the name of the application / watch face
* The folder MUST contain an app.py python file containing the application / watch face logic
* The folder MAY contain an icon.png file that will be used as the icon for the application if present
* The folder MAY contain a screenshot.png file that will be used as the screenshot of the application in the documentation.


GaryM0101:
With these changes, have you tried building/testing Wasp-os from your fork? I can see already that path changes would be needed for the wasp.toml initially. I think that other configuration files like the manifest might be okay since you're not separating the "built-in" apps. When things smooth out, I'll clone your fork, compile and give it a try.

GaryM0101:
Also when you say "MUST contain an app.py python file", I don't see how this would work. Your file naming "app.py" for every application is not compatible with wasp. See https://wasp-os.readthedocs.io/en/latest/appguide.html#id15

BeatLink:
Right, the docs would need to be changed. But I tried running the apps with make sim and everything worked okay

fgaz:
> But I tried running the apps with make sim and everything worked okay

I think this is because the old structure is still in place. Instead of copying files from the old locations, try moving them. Then check that everything works by running `make dist apps check`.

You also need to sign-off your commits (see the details of the failing DCO check).

fgaz:
Regarding the design, it looks nice, but... does any app in the project actually follow these guidelines? To me it looks like the design of most apps was guided by practical concerns driven by the small screen size.

The infinitime docs have some [practical ui guidelines](https://github.com/InfiniTimeOrg/InfiniTime/blob/main/doc/ui_guidelines.md), maybe we could adapt / link to them.

daniel-thompson:
> Regarding the design, it looks nice, but... does any app in the project actually follow these guidelines?
> To me it looks like the design of most apps was guided by practical concerns driven by the small screen size.

So, I guess I did most of the widget design so I can answer that.

The practical concerns that the current UI is built around is not simply about landing big fingers on a small watch. There is also the cost of drawing the widgets: on  PineTime the RAM is small, MicroPython isn't very fast and the bandwidth to shift pixels between the CPU and the screen is shockly bad (in the best case it has pixel rate is 500KHz and most of the time the command overhead makes it much lower).

So the reason most of the core UI (watch, launcher, step counter, stopwatch, etc) uses a borderless design is partly for performance (cheap to draw) and partly because I felt a borderless design made the watch "feel" more premium than it actually is (because it hides where the relatively large bezel actually is).

BeatLink:
Ahh okay. I guess i can close this then

BeatLink:
Could the design be more standardized however? Like using symbolic icons and common colours

Issue #222, opened 2021-06-23
Stuck in bootloader after initial flashing (P8)


arvchristos:
Hi,
I followed the documentation and flashed my P8b (MOY-TON5). After flashing `micropython.py`, instead of being greeted with the wasp-os watchface, I am stuck in the bootloader (Start with bluetooth icon). After I press the button, I see the following pattern:

> Wasp-os logo for 2 seconds, black screen of 5 seconds,Wasp-os logo for 2 seconds (again), <- Start and the BT logo appear

So far, I have tried the following: 
* removing `boot.py` from the manifest
* building and flashing both the latest release as well as the master branch tip.
* Reflashing the bootloader using `reload.zip`
* Commenting out StepCounter from register_defaults

All of the above did not change the aforementioned pattern. Any ideas?

Pittconnect:
Flash display or stuck in a cycle has happened before and it called softbricking the device. where the swatch being very slow in processing and responding to the program command request compared with your phone people often press the button prematurely thick can softbrick the device and set it into a loop of flickering display or switching between two screens. Solution is wait for the battery to drain completely and upload the new firmware here are different methods https://wiki.pine64.org/wiki/Switching_your_PineTime_between_InfiniTime_and_Wasp-os But you may also try recovery bootloader https://github.com/Peetz0r/wasp-reloader/releases/tag/infinitime-0.14.1-recovery

evanhuang117:
> Flash display or stuck in a cycle has happened before and it called softbricking the device. where the swatch being very slow in processing and responding to the program command request compared with your phone people often press the button prematurely thick can softbrick the device and set it into a loop of flickering display or switching between two screens. Solution is wait for the battery to drain completely and upload the new firmware here are different methods https://wiki.pine64.org/wiki/Switching_your_PineTime_between_InfiniTime_and_Wasp-os But you may also try recovery bootloader https://github.com/Peetz0r/wasp-reloader/releases/tag/infinitime-0.14.1-recovery

I'm also having this problem and flashed the MCUBoot bootloader with SWD and it does not boot, completely black screen. What do you think I should do?

daniel-thompson:
Thanks for the comments although I think there might have been a bit of a misunderstanding here. The original reporter is describing a problem using the P8b rather than the PineTime. AFAIK neither Infinitime nor mcuboot have been ported to the P8b.

[...]

Issue #82, opened 2020-10-16
Gatttool and hcitool depreciated from arch


Izaic:
Gatttool and hcitool have been dropped from arch's bluez-utils package for a long time.

Would it be possible to replace the dependency? Is there even any distros with these tools packaged still, or do you have to manually build them?

daniel-thompson:
This is an inherited dependency.

If you just want to raise a ticket then please add it against the correct component (either https://github.com/dingari/ota-dfu-python or https://github.com/daniel-thompson/ota-dfu-python ). That is also the right place to send pull requests.

Alternatively if you aware an an alternative dfu implementation that wasp-os could use instead then please let me know.

fgaz:
Your fork doesn't have issues enabled, is there another place where I can open a ticket?

PR #523, opened 2025-04-14
fix: Update Dockerfile and its build process

6arms1leg:
This PR updates the developer environment's Dockerfile and its build process:

* Add Make targets for the Docker-related entry points (build, push, run)
* Use Docker Compose to handle the build process
* Clean up the Dockerfile
* Update the Docker image to Ubuntu 24
* Add Bluetooth support to the Docker image

I have a Colmi P8b and had quite a bit of trouble building and flashing Wasp OS (mainly #222).
Using the previous Docker image, the compiled/linked Wasp OS binary wouldn't run and flashing via Bluetooth from within the Docker container didn't work.
Flashing on GNU/Linux via Bluetooth without Docker didn't work either (#82).

With this Docker image, I am able to compile/link Wasp OS and flash it via Bluetooth (all inside the Docker container).


Makefile
```diff
@@ -170,3 +170,32 @@ dist: docs
...
 	tar -C .. -zcf $(DIST).tar.gz $(notdir $(DIST))
 	(cd ..; zip -9r $(DIST).zip $(notdir $(DIST)))
+
+build-docker-image:
+	docker compose \
+	    --file ./tools/docker/docker-compose-build.yml \
+	    build \
+	        --pull
+
+# TODO: Create Wasp OS DockerHub account?
```
daniel-thompson, line 180: I think it might be better to push to the github container repository instead. You're certainly welcome to explore whether the github CI bits 'n pieces can be hooked up to push this automatically when anything in tools/docker is changed.

However, whether you pick that work up or not, I don't think it merits a TODO in the source for this.

daniel-thompson (changes requested):
Thanks for the changes.

It looks like these changes would require the docs to be updated though: https://github.com/wasp-os/wasp-os/blob/master/docs/install.rst#install-prerequisites-via-docker

Makefile
6arms1leg: I removed the `TODO:` inline comment.

6arms1leg:
> Thanks for the changes.
> 
> It looks like these changes would require the docs to be updated though: https://github.com/wasp-os/wasp-os/blob/master/docs/install.rst#install-prerequisites-via-docker

I updated the documentation accordingly.

daniel-thompson (approved):

